PR #50, opened 2023-03-28
Add generator support for optional parameters

SlyryD:
- Add generator support for optional parameters
- Use specific Microsoft.SemanticKernel version so it matches between the generator command in README.md and the csproj
- Fix variable name typo
- Allow Node 16+

### DIFF
--- a/examples/semantic-kernel/semantic-kernel.csproj
+++ b/examples/semantic-kernel/semantic-kernel.csproj
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SemanticKernel" Version="*-*" />
+    <PackageReference Include="Microsoft.SemanticKernel" Version="0.9.61.1-preview" />
   </ItemGroup>
 
 </Project>
--- a/src/NodeApi.Generator/TypeDefinitionsGenerator.cs
+++ b/src/NodeApi.Generator/TypeDefinitionsGenerator.cs
@@ -656,7 +656,8 @@ public class TypeDefinitionsGenerator : SourceGenerator
             }
             else
             {
-                return $"{TSIdentifier(parameters[0].Name)}: {parameterType}";
+                string optionalToken = parameters[0].IsOptional ? "?" : "";
+                return $"{TSIdentifier(parameters[0].Name)}{optionalToken}: {parameterType}";
             }
         }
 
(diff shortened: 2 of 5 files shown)


vmoroz (approved):
LGTM

src/NodeApi.Generator/TypeDefinitionsGenerator.cs
```diff
@@ -666,7 +666,8 @@ private string GetTSParameters(ParameterInfo[] parameters)
         foreach (ParameterInfo p in parameters)
         {
             string parameterType = GetTSType(p);
-            s.AppendLine($"{TSIdentifier(p.Name)}: {parameterType},");
+            string optionalToken = p.IsOptional ? "?" : "";
+            s.AppendLine($"{TSIdentifier(p.Name)}{optionalToken}: {parameterType},");
```
jasongin, line 670: This change is needed also at line 659 above. That formats a single parameter on the same line as the method, while multiple parameters here are formatted each on separate lines.

SlyryD: Fixed! A lot of parameters are coming through like `orgId?: string | null`. Is that how we want to handle optional and nullable?

jasongin: Probably not. I think optional parameters should just use `?` without the `| null`. Then if/when JS passes an `undefined` value (or omits the argument) it gets implicitly converted to `null` in .NET.

See also [the issue tracking null/undefined confusion](https://github.com/microsoft/node-api-dotnet/issues/58). Early in the project we were unsure (and debated) how to handle some things. Now that it's further along I think we can do a more complete analysis and decide on some consistent patterns.

examples/semantic-kernel/semantic-kernel.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SemanticKernel" Version="*-*" />
+    <PackageReference Include="Microsoft.SemanticKernel" Version="0.8.56.1-preview" />
```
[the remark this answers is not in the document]
SlyryD: Fixed!

src/NodeApi.Generator/TypeDefinitionsGenerator.cs
jasongin: The `| null` issue can be addressed separately, so I'll resolve this issue and merge the PR.

PR #51, opened 2023-03-28
.NET Framework 4.7.2 support

jasongin:
:warning: **DRAFT STATUS** :warning:

At this point the tests are passing against .NET Framework 4.7.2, and the `dotnet-module` example works. This still needs a little more clean-up, to fix formatting and other issues.

 - Add `net472` build target.
   - Several dependencies require at least .NET 4.7.2, so I didn't try to go lower than that for now.
   - Add conditional references to NuGet packages that fill in some APIs not built in to .NET 4.x:
       - `System.Memory`
       - `Microsoft.Bcl.AsyncInterfaces`
       - `Nullability.Source`
    - Define `NetFramework` build variable and `NETFRAMEWORK` preprocessor constant when building the `net472` target.
    - Use `#ifdef NETFRAMEWORK` in many places where there could be alternative (often less efficient) code that is compatible with .NET Framework.
        - Exclude anything referencing `IReadOnlySet<>`, which doesn't exist in .NET Framework.
    - Use `Marshal.GetFunctionPointerForDelegate()` instead of unmanaged delegates.
        - Updated to preserve use of unmanaged delegates for .NET 6+.
    - Add empty definitions for `[NotNull]` and similar control-flow analysis attributes that don't exist in .NET Framework.
  - Implement .NET Framework loading in `NativeHost` using `mscoree` APIs.
     - These APIs are COM-like therefore tedious to P/Invoke, but it works.
  - Update npm packaging script to bundle additional assemblies required by .NET Framework, like `System.Memory`.
  - Update `node-api-dotnet` npm package initialization JS code to support requiring a specific .NET version:
     - `require('node-api-dotnet')` loads the default version: .NET 6
     - `require('node-api-dotnet/net472')` or `require('node-api-dotnet/net7.0')` loads the specific version of .NET

Fixes: #53


src/NodeApi/Native/JSNativeApi.cs
```diff
@@ -832,19 +832,39 @@ public static void SetObjectTypeTag(this JSValue thisValue, in napi_type_tag typ
 
     public static unsafe void SetObjectTypeTag(this JSValue thisValue, ref Guid typeGuid)
     {
+        #if NETFRAMEWORK
+        byte[] guidBytes = typeGuid.ToByteArray();
```
vmoroz, line 836: >ToByteArray [](http://example.com/codeflow?start=36&length=11)

It would be better to avoid memory allocation here.
At some point I was creating struct as a local variable and was copying Guid bytes there. While that code cannot be used here "as-is", but just in case how it looked like: [https://github.com/vmoroz/napi-cs/blob/9c216ef0626b07fbfcd4433ff8d527a3ac830504/NodeApi/JSNative.cs#L844](https://github.com/vmoroz/napi-cs/blob/9c216ef0626b07fbfcd4433ff8d527a3ac830504/NodeApi/JSNative.cs#L844)

src/NodeApi/JSProxy.cs
```diff
@@ -120,6 +120,21 @@ public Handler(string? name = null)
...
+        public GetOwnPropertyDescriptor? GetOwnPropertyDescriptor { get; set; }
+        public GetPrototypeOf? GetPrototypeOf { get; set; }
+        public Has? Has { get; set; }
+        public IsExtensible? IsExtensible { get; set; }
+        public OwnKeys? OwnKeys { get; set; }
+        public PreventExtensions? PreventExtensions { get; set; }
+        public Set? Set { get; set; }
+        public SetPrototypeOf? SetPrototypeOf { get; set; }
+#else
         public Apply? Apply { get; init; }
```
vmoroz, line 138: >init [](http://example.com/codeflow?start=35&length=4)

We should also just add missing attributes? I wonder if we should be using the PolySharp project as a dependency. I guess it adds all such attributes for us.

jasongin:
Update: Tests are all passing against `net472` target.

I skipped the `multi-instance` test for `net472` for now. I started to implemented that using .NET Framework `AppDomain`, then I remembered how difficult app domains are to work with, and decided that isn't critical right now; we can add it later.

Now the builds fail just from `dotnet format` issues. Some of those are for suggestions that don't work with .NET Framework, for example using method overloads that take `Span<>` instead of `byte[]`. I plan to disable those analyzer rules.

And, I still want to do another code-cleanup pass, and address PR comments.

src/NodeApi/Native/JSNativeApi.cs
jasongin: `MemoryMarshal.CreateReadOnlySpan()` is not available in .NET Framework.

But this works:
```C#
        Guid guid = typeGuid;
        napi_type_tag typeTag;
        long* pGuid = (long*)&guid;
        long* pTag = (long*)&typeTag;
        pTag[0] = pGuid[0];
        pTag[1] = pGuid[1];
```

src/NodeApi/JSProxy.cs
jasongin: I did add some missing attributes... but the issue with `{ get; init; }` here is unrelated to attributes. .NET Framework runtime does not support it as far as I can see. The error is:

> CS0518 Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported

vmoroz: Here they say that we can just define the attribute ourselves to make it pass: https://youtu.be/WjTifkBlJHQ?t=170

jasongin: OK, it works. I was confused because it is not an **attribute**.

jasongin:
PR builds are all green... which is surprising because I'm still hitting a `dotnet format` bug locally:

> dotnet format --no-restore --severity info --verify-no-changes
C:\Users\jasongin\.nuget\packages\nullability.source\2.1.0\contentFiles\cs\netstandard2.0\Nullability.Source\NullabilityInfoExtensions.cs(128,60): info IDE0060: Remove unused parameter 'name' [D:\node-api-dotnet\src\NodeApi.Generator\NodeApi.Generator.csproj]

That nuget package actually contributes source files instead of a DLL. Then `dotnet format` reports an issue against it -- and what's worse, _it cannot be suppressed via `.editorconfig`_. (I tried many variations.)

This is far from the first time I've seen differences between local and online `dotnet format` runs; it has been rather annoying. But since it's not blocking the PR builds, I'm not going to block this PR on it.

So, @vmoroz  I think this is ready to merge, unless there are any further review comments?

vmoroz (approved):
LGTM

PR #68, opened 2023-03-29
Build improvements

jasongin:
This change simplifies build steps for example projects, and improves how the test runner automatically builds dependencies.

 - Automatically run an MSBuild child task to publish the native host (for `net7.0` target framework) before packing the managed host npm package.
 - Remove the `dotnet publish` step from a few places in documentation, since it no longer needs to be called separately before `dotnet pack`.
 - Change the test runner to invoke `dotnet build` as a child process rather than driving MSBuild in-proc, when building dependencies.
 - Update to the current version of the `Microsoft.CodeAnalysis.CSharp` nuget package. The previous referenced version had a bug that broke MSBuild up-to-date checks.

The test runner change enables it to build dependencies using a different target framework than what the tests are currently targeting. It will be critical for .NET Framework 4.x testing, because that will still depend on the AOT-compiled .NET 7 native host. (#51 testing was blocked by this issue.)

### DIFF
--- a/src/NodeApi.DotNetHost/NodeApi.DotNetHost.csproj
+++ b/src/NodeApi.DotNetHost/NodeApi.DotNetHost.csproj
@@ -12,6 +12,14 @@
     <ProjectReference Include="..\NodeApi\NodeApi.csproj" GeneratePathProperty="true" PrivateAssets="all" />
   </ItemGroup>
 
+  <Target Name="PublishNativeHost" BeforeTargets="Pack">
+    <MSBuild
+      Projects="..\NodeApi\NodeApi.csproj"
+      Targets="Publish"
+      Properties="Configuration=$(Configuration);TargetFramework=net7.0"
+    />
+  </Target>
+
   <Target Name="NpmPack"
     AfterTargets="Pack"
     Outputs="$(PackageOutputPath)node-api-dotnet-$(PackageVersion).tgz"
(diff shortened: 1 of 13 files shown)


src/NodeApi.DotNetHost/NodeApi.DotNetHost.csproj
```diff
@@ -12,6 +12,14 @@
     <ProjectReference Include="..\NodeApi\NodeApi.csproj" GeneratePathProperty="true" PrivateAssets="all" />
   </ItemGroup>
 
+  <Target Name="PublishNativeHost" BeforeTargets="Pack">
```
vmoroz, line 15: ><Target Name="PublishNativeHost" BeforeTargets="Pack"> [](http://example.com/codeflow?start=2&length=54)

How it would work in the CI test I am working on?
Currently I publish in different environments and then collect the binaries to pack them in the single job. Binding publishing and packing makes it impossible to do. What is the new proposed algorithm?

vmoroz (approved):
:shipit:

src/NodeApi.DotNetHost/NodeApi.DotNetHost.csproj
jasongin: I added `Condition=" '$(NoPublish)' != 'true' "` here. So you can skip the auto-publish like this:
```
dotnet pack -p:NoPublish=true
```

PR #75, opened 2023-04-02
Support JS impl of .NET interface for dynamic mode

jasongin:
So far, this project has enabled implementing a .NET interface in JavaScript, when the interface is exported by a .NET Node API module. That relied on compile-time source-generation to define a proxy .NET class that implements the .NET interface by forwarding the calls to JS.

With this change it is now possible to do the same with dynamic invocation. This uses `Reflection.Emit` to dynamically define the same kind of interface proxy classes. It is a little more complex than the source generation, since another level of indirection is required to hook up the marshalling expressions/delegates. And emitting IL is tedious and verbose. (I prefer to use compiled lambda expressions instead of emitting IL when possible, but it's not possible to do that directly in this case.)

### DIFF
--- a/src/NodeApi.DotNetHost/AssemblyExporter.cs
+++ b/src/NodeApi.DotNetHost/AssemblyExporter.cs
@@ -243,14 +243,16 @@ internal class AssemblyExporter
         {
             if (member is PropertyInfo property &&
                 property.PropertyType.Assembly == type.Assembly &&
-                IsSupportedType(property.PropertyType))
+                IsSupportedType(property.PropertyType) &&
+                !JSMarshaller.IsConvertedType(property.PropertyType))
             {
                 ExportClass(property.PropertyType);
             }
             else if (member is MethodInfo method &&
                 IsSupportedMethod(method) &&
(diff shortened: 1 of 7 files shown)


vmoroz:
I just wonder which constructs make it impossible to use the Expression-based code gen.

vmoroz (approved):
:shipit:

jasongin:
@vmoroz
> I just wonder which constructs make it impossible to use the Expression-based code gen.

1. There's no way to dynamically construct types (classes, properties, methods, etc) using only expressions. It has to be done with `TypeBuilder`, `MethodBuilder`, and related APIs.
2. `MethodBuilder` and similar do not provide a way to assign a delegate or lambda expression as the implementation. The only way to implement a method of a dynamically generated type is by [emitting IL](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.emit.methodbuilder.getilgenerator).
3. The bulk of the implementation is still using the generated expression trees. It emits just enough IL instructions to invoke delegates generated from expression trees.

PR #77, opened 2023-04-02
Avoid dependency on Nerdbank.GitVersioning in published Nuget

vmoroz:
Currently our published Nuget packages depend on the Nerdbank.GitVersioning.
This is not needed, and this PR removes this dependency.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -37,7 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Nerdbank.GitVersioning" PrivateAssets="none" />
+    <PackageReference Include="Nerdbank.GitVersioning" PrivateAssets="all" />
   </ItemGroup>
 
   <Import Project="./rid.props" />


Directory.Build.props
```diff
@@ -37,7 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Nerdbank.GitVersioning" PrivateAssets="none" />
+    <PackageReference Include="Nerdbank.GitVersioning" PrivateAssets="all" />
```
jasongin, line 40: Does this cause the `Nerdbank.GitVersioning` assembly to be copied to the project output directory?

jasongin (approved):

Directory.Build.props
vmoroz: No, I do not see it being copied. But it avoids the dependency entries in the *.nuspec files.
Using `PrivateAssets="all"` is recommend by the [`Nerdbank.GitVersioning` docs](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/msbuild.md).

Issue #55, opened 2023-03-29
Separate contexts for multiple .NET modules


jasongin:
Currently if more than one .NET module is loaded, things get mixed up because the "module context" is overwritten by each loaded module. We need to create something like a _virtual_ module context that is distinct for each .NET module (while they must all share the single Node API module context).

PR #81, opened 2023-04-03
Refactor context to support multiple modules

jasongin:
As described in #55, there were problems with loading multiple (non-AOT) modules in the same process. Module-level methods could get the wrong `this` pointer (likely causing a cast exception) because each module overwrites the reference to the module object.

 - Add a new `JSModuleContext` that is specific to each module, while the existing `JSContext` is shared across all .NET modules loaded by the same host.
 - Add a `JSValueScopeType.Module` that is the scope type for module initialization. It inherits the shared `JSContext` from the host but creates a new `JSModuleContext` instance.
 - Replace the single `JSValueScope.ModuleContext` property with distinct `Context` and `ModuleContext`.
 - Pass the module context to callbacks via the descriptor: - either `JSPropertyDescriptor` or `JSCallbackDescriptor`. Then assign it to the `JSValueScope` that is initialized for the callback.

We could consider moving some additional things from `JSContext` to `JSModuleContext`, though at this point it's not obvious to me that anything needs to move: if multiple modules are passing around objects among them, then it probably makes sense to have shared object and reference tracking.

There are some small additional fixes included here:
 - Don't use `global` for passing `require()` in when initializing a .NET hosted module. (Though it is still necessary for AOT.)
 - Fix an occasional hang in the test runner when running tests.

Fixes: #55

### DIFF
--- a/src/NodeApi.Generator/ModuleGenerator.cs
+++ b/src/NodeApi.Generator/ModuleGenerator.cs
@@ -118,7 +118,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
                         "Module class must have public visibility.");
                 }
 
-                // TODO: Check for a public constructor that takes a single JSContext argument.
+                // TODO: Check for a public constructor that takes a single JSRuntimeContext argument.
 
                 moduleInitializers.Add(type);
             }
@@ -297,7 +297,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
         s += "using var scope = new JSValueScope(JSValueScopeType.Module, env);";
         s += "try";
         s += "{";
-        s += "JSContext context = scope.Context;";
+        s += "JSRuntimeContext context = scope.RuntimeContext;";
         s += "JSValue exportsValue = new(exports, scope);";
         s++;
 
@@ -380,7 +380,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
         }
         else
         {
-            s += $"exportsValue = new JSModuleBuilder<JSContext>()";
+            s += $"exportsValue = new JSModuleBuilder<JSRuntimeContext>()";
             s.IncreaseIndent();
         }
 
--- a/src/NodeApi.Generator/ModuleGenerator.cs
+++ b/src/NodeApi.Generator/ModuleGenerator.cs
@@ -294,10 +294,10 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
         // The main initialization entrypoint is called by the `ManagedHost`, and by the unmanaged entrypoint.
         s += $"public static napi_value {ModuleInitializeMethodName}(napi_env env, napi_value exports)";
         s += "{";
-        s += "using var scope = new JSValueScope(JSValueScopeType.Root, env);";
+        s += "using var scope = new JSValueScope(JSValueScopeType.Module, env);";
         s += "try";
         s += "{";
-        s += "JSContext context = scope.ModuleContext;";
+        s += "JSContext context = scope.Context;";
         s += "JSValue exportsValue = new(exports, scope);";
         s++;
 
@@ -491,17 +491,9 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
         if (moduleType != null)
         {
             // Construct an instance of the custom module class when the module is initialized.
-            // The module class constructor may optionally take a JSContext parameter. If an
-            // appropriate constructor is not present then the generated code will not compile.
-            IEnumerable<IMethodSymbol> constructors = moduleType.GetMembers()
-                .OfType<IMethodSymbol>().Where((m) => m.MethodKind == MethodKind.Constructor);
-            IMethodSymbol? constructor = constructors.SingleOrDefault((c) =>
-                c.Parameters.Length == 1 && c.Parameters[0].Type.Name == "JSContext") ??
-                constructors.SingleOrDefault((c) => c.Parameters.Length == 0);
-            string contextParameter = constructor?.Parameters.Length == 1 ?
-                "context" : string.Empty;
+            // IF a no-args constructor is not present then the generated code will not compile.
             string ns = GetNamespace(moduleType);
-            s += $".ExportModule(new {ns}.{moduleType.Name}({contextParameter}), (JSObject)exportsValue);";
+            s += $".ExportModule(new {ns}.{moduleType.Name}(), (JSObject)exportsValue);";
         }
         else
         {
--- a/src/NodeApi.Generator/ModuleGenerator.cs
+++ b/src/NodeApi.Generator/ModuleGenerator.cs
@@ -491,7 +491,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
         if (moduleType != null)
         {
             // Construct an instance of the custom module class when the module is initialized.
-            // IF a no-args constructor is not present then the generated code will not compile.
+            // If a no-args constructor is not present then the generated code will not compile.
             string ns = GetNamespace(moduleType);
             s += $".ExportModule(new {ns}.{moduleType.Name}(), (JSObject)exportsValue);";
         }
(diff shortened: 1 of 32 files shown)


vmoroz:
We were intending to rename `JSContext` to `JSModuleContext` at some point.
The `JSContext` name is too generic.
I guess based on the new semantics we should consider renaming it to the `JSRuntimeContext`, `JSNativeContext`, or `JSNativeModuleContext`/`JSModuleNativeContext`.

src/NodeApi/Interop/JSModuleContext.cs
```diff
@@ -0,0 +1,52 @@
...
+        if (IsDisposed) return;
+
+        IsDisposed = true;
+
+        if (Module is IDisposable module)
+        {
+            module.Dispose();
+        }
+
+        GC.SuppressFinalize(this);
```
vmoroz, line 50: >GC.SuppressFinalize(this); [](http://example.com/codeflow?start=8&length=26)

I wonder if it is needed when we have no class Finalizer.

test/TestCases/common/index.js
```diff
@@ -118,18 +118,21 @@ exports.dotnetModule = process.env.TEST_DOTNET_MODULE_PATH;
...
-      .initialize(exports.dotnetVersion, exports.dotnetHost.replace(/\.node$/, '.DotNetHost.dll'))
+      .initialize(
+        exports.dotnetVersion,
+        exports.dotnetHost.replace(/\.node$/, '.DotNetHost.dll'),
+        require)
       .require(exports.dotnetModule);
   } else {
+    // The Node API module may need the require() function at initialization time; passing it as a
+    // global is the only solution for AOT, since it cannot be obtained via any `napi_*` function.
+    global.require = require;
```
vmoroz, line 134: >require [](http://example.com/codeflow?start=11&length=7)

Can we use a different name on the global that pseudo-scopes it to our project?
E.g. __node_api_dotnet_require.

test/TestBuilder.cs
```diff
@@ -343,7 +344,15 @@ private static bool GetNoBuild()
         process.BeginOutputReadLine();
         process.BeginErrorReadLine();
 
-        process.WaitForExit();
+        // Process.WaitForExit() may hang when redirecting output because it actually waits for the
```
vmoroz, line 347: >// [](http://example.com/codeflow?start=8&length=2)

This is a great finding! Lately our Ubuntu builds started to hang all the time.
I believe it happened after the locks were removed from the result reporting. I was wondering if the cause was the corruption of logWriter caused by accessing from multiple threads.

src/NodeApi.Generator/ModuleGenerator.cs
```diff
@@ -491,17 +491,9 @@ private IEnumerable<ISymbol> GetModuleExportItems()
...
-            // The module class constructor may optionally take a JSContext parameter. If an
-            // appropriate constructor is not present then the generated code will not compile.
-            IEnumerable<IMethodSymbol> constructors = moduleType.GetMembers()
-                .OfType<IMethodSymbol>().Where((m) => m.MethodKind == MethodKind.Constructor);
-            IMethodSymbol? constructor = constructors.SingleOrDefault((c) =>
-                c.Parameters.Length == 1 && c.Parameters[0].Type.Name == "JSContext") ??
-                constructors.SingleOrDefault((c) => c.Parameters.Length == 0);
-            string contextParameter = constructor?.Parameters.Length == 1 ?
-                "context" : string.Empty;
+            // IF a no-args constructor is not present then the generated code will not compile.
```
vmoroz, line 494: >IF [](http://example.com/codeflow?start=15&length=2)

nit: "If"

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -158,6 +163,8 @@ private JSValue InitializeFrameworkHost(Version minVersion, string managedHostPa
 
             // Create an "exports" object for the managed host module initialization.
             JSValue exportsValue = JSValue.CreateObject();
+            exportsValue.SetProperty("require", require);
```
vmoroz, line 166: >exportsValue.SetProperty("require", require); [](http://example.com/codeflow?start=12&length=45)

Should we rather define a managed entry point and pass it as a normal parameter. It feels wrong that we add a new property to `exports`.

src/NodeApi/Interop/JSCallbackDescriptor.cs
```diff
@@ -25,7 +31,18 @@ namespace Microsoft.JavaScript.NodeApi.Interop;
 
     public JSCallbackDescriptor(JSCallback callback, object? data = null)
     {
-        Callback = callback;
+        JSValueScope currentScope = JSValueScope.Current ??
+            throw new InvalidOperationException("No current scope.");
```
vmoroz, line 35: Should we rather add a new property to `JSValueContext` that would do the verification. I believe we do this check in multiple places.

vmoroz (approved):
:shipit:

test/TestBuilder.cs
jasongin: I had removed the locks because I thought maybe they were contributing to the hang. That didn't help, though I didn't think it made anything worse. Anyway it should be good now.

jasongin:
> I guess based on the new semantics we should consider renaming it to the `JSRuntimeContext`, `JSNativeContext`, `JSNativeModuleContext`/`JSModuleNativeContext`, or `JSHostContext`.

"Native" sounds wrong since it is used also by "managed" .NET modules.

`JSHostContext` could be OK, though it might sound like it is only for .NET code that is hosting JavaScript, when it may be the other way around.

So I like `JSRuntimeContext` best. I'll rename to that unless you have any objections.

src/NodeApi/Interop/JSModuleContext.cs
jasongin: The doc says it has no effect on an object that does not have a finalizer.

We have it in ~10 places in the codebase, but I think none of those are needed except for `JSReference` which does define a finalizer.

src/NodeApi/DotNetHost/NativeHost.cs
jasongin: It gets overwritten by the `require()` method defined by the `ManagedHost`. So this property is never visible to applications.

test/TestCases/common/index.js
jasongin: I'm not sure that would be better. This actually has to be set by user code at the moment, so it might be a little strange if user code had to set a property with that scoped name.

Later we might want to have the TS generator also output a `.js` wrapper for the AOT `.node` module. That would take care of both referencing the `.d.ts` and setting this global.

src/NodeApi/Interop/JSCallbackDescriptor.cs
jasongin: I changed the existing `JSValueContext.Current` to throw if null. It works well, since it really never should be null.

src/NodeApi/Interop/JSModuleContext.cs
jasongin: Cleaned up.

test/TestCases/common/index.js
vmoroz: The problem with the global name `require` is that it will affect all modules in a Node.JS applications.
In some cases, in may cause some unexpected behaviors that will be in the end point as the root cause to our code and security experts will mark our NPM package as unsecure.
It must be any other name that must not cause side effects in other modules.
Any global name is callable from any module as it was locally defined in that module.
Having a special name that is visually related to our module is very important in such scenario.

vmoroz: BTW, we had discussed the issue of passing `module` parameter to the new entry point and generally it was not rejected. It was considered that it could be part of the new Node-API version. Though, it may take a year or more before we could use it.

vmoroz: What if we create a special node-api-dotnet global property object and the `require` will be just its property?
E.g. `global.node_api_dotnet = { require };`

vmoroz: Another alternative is to make sure that we always delete this property inside of the entry point function.
This way, I agree with you, it doesn't matter how we call it.
My main concern is that we must not leak this global property to anyone else.

jasongin: Updated to use `global.node_api_dotnet = { require }`.

> Another alternative is to make sure that we always delete this property inside of the entry point function.

I wouldn't want to _delete_ `global.require`, as some other code may rely on it. It is somewhat likely for applications or frameworks to set `global.require = require`, because Node.js itself does that in some contexts. For example:
```
$ node -p "typeof global.require"
function
```

PR #82, opened 2023-04-03
Fix test-results reporting for PR builds from repo forks

### LINKED ISSUE #71: Fix test-results reporting for PR builds from repo forks
Currently PR builds appear to fail for any PR from a fork of the project repo, because the "Publish test results" step fails. The basic test results are still visible in the build logs, but the nice HTML results report doesn't succeed.

A solution is described at https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories

vmoroz:
Resolves #71
The PR follows the solution described here: https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories
As a part of this solution, we split up the test reporting into a separate workflow.
The build script is tested against a fork repo.

Related changes:
- Added `configuration` to the build matrix
- `runner.os` is replaced by `matrix.os` to simplify matching attribute names.
- Renamed the build workflow to 'PR Verification'.
- Do not start build workflow on push.
- Execute all tests as long as 'pack' succeeded regardless of other tests failure.
- Do not execute the most steps when build is cancelled.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 # https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net
 
-name: napi-dotnet build and test
+name: PR Verification
 
 on:
   push:
@@ -16,14 +16,14 @@ permissions:
 
 jobs:
   build:
-
-    runs-on: ${{ matrix.os }}
-
     strategy:
       matrix:
         os: [ windows-latest, macos-latest, ubuntu-latest ]
         node-version: [ 18.x ]
+        configuration: [ Release ]
       fail-fast: false  # Don't cancel other jobs when one job fails
+
+    runs-on: ${{ matrix.os }}
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -73,37 +73,37 @@ jobs:
...
+        --results-directory "test-dotnet7-node${{ matrix.node-version }}-${{ matrix.configuration }}"
 
     - name: Upload test logs
       uses: actions/upload-artifact@v3
       with:
-        name: test-logs-${{ runner.os }}-node${{ matrix.node-version }}
-        path: out/obj/Release/**/*.log
+        name: test-logs-${{ matrix.os }}-node${{ matrix.node-version }}-${{ matrix.configuration }}
+        path: out/obj/${{ matrix.configuration }}/**/*.log
       if: ${{ always() }}
```
jasongin, line 105: Currently if tests for one of the target frameworks fails, then the following test steps do not execute. I wonder if there is a condition such that we can run tests on successful build, but regardless of success or failure of earlier test steps?

jasongin (approved):

.github/workflows/build.yml
vmoroz: Agree, I was looking how to address it too. I am going to augment the PR if I find the fix.
(I found a way to test it in my fork and the code is not ready yet.)

vmoroz: The latest iteration addresses it.

PR #84, opened 2023-04-05
Extract NativeLibrary polyfills to another class

jasongin:
This makes it easier to use `NativeLibrary` methods from other classes and assemblies. We'll need it for hosting JS engines.


vmoroz (approved):
:shipit:

PR #89, opened 2023-04-09
Implement an example that uses Hermes engine.

vmoroz:
This is a new example that uses Hermes JavaScript engine created by Meta.
The hermes.dll from the forked hermes-windows project exports Node-API as an ABI-safe API.
The new hermes-example shows how it can be used.

The hermes.dll does not implement Node.JS-specific APIs from `node_api.h`.
In this PR we detect if the thread-safe functions are available, and if they are not, then we use the new `JSDispatcherQueue` for the `SynchronizationContext` implementation. The `JSDispatcherQueue` API design is "inspired" by the WinRT [`DispatcherQueue`](https://learn.microsoft.com/en-us/uwp/api/windows.system.dispatcherqueue?view=winrt-22621).

The example is very basic, but it works.
We can improve it or make it more interesting in the follow up PRs.

Side note: this PR is submitted from my fork. We should see if the new PR pipeline handles it correctly.

### DIFF
--- a/examples/hermes-engine/Program.cs
+++ b/examples/hermes-engine/Program.cs
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Hermes.Example;
+using Microsoft.JavaScript.NodeApi;
+using static Microsoft.JavaScript.NodeApi.JSNativeApi.Interop;
+
+JSDispatcherQueueController controller = JSDispatcherQueueController.CreateOnDedicatedThread();
+controller.DispatcherQueue.TryEnqueue(() =>
+{
+    HermesApi.Load("hermes.dll");
+    using var runtime = new HermesRuntime();
+    using var scope = new JSValueScope(JSValueScopeType.Root, (napi_env)runtime);
+
+    JSNativeApi.RunScript("x = 2");
+    Console.WriteLine($"Result: {(int)JSValue.Global["x"]}");
+});
+
+controller.DispatcherQueue.TryEnqueue(() =>
+{
+    controller.ShutdownQueueAsync();
+});
--- a/src/NodeApi/Interop/JSSynchronizationContext.cs
+++ b/src/NodeApi/Interop/JSSynchronizationContext.cs
@@ -3,19 +3,51 @@
 
 using System;
 using System.Threading;
+using Hermes.Example;
 
 namespace Microsoft.JavaScript.NodeApi.Interop;
 
-public sealed class JSSynchronizationContext : SynchronizationContext, IDisposable
+public class JSSynchronizationContext : SynchronizationContext, IDisposable
 {
-    private readonly JSThreadSafeFunction _tsfn;
-
     public bool IsDisposed { get; private set; }
 
     public static new JSSynchronizationContext? Current
         => SynchronizationContext.Current as JSSynchronizationContext;
 
-    public JSSynchronizationContext()
+    public static JSSynchronizationContext Create()
+    {
+        if (JSThreadSafeFunction.IsAvailable)
+        {
+            return new JSTsfnSynchronizationContext();
+        }
+        else if (JSDispatcherQueue.GetForCurrentThread() is JSDispatcherQueue queue)
+        {
+            return new JSDispatcherSynchronizationContext(queue);
+        }
+        else
+        {
+            throw new JSException("Cannot create synchronization context.");
+        }
+    }
+
+    protected JSSynchronizationContext() { }
+
+    public virtual void Dispose()
+    {
+        if (IsDisposed) return;
+        IsDisposed = true;
+    }
+
+    public virtual void OpenAsyncScope() { }
+
+    public virtual void CloseAsyncScope() { }
+}
+
+public sealed class JSTsfnSynchronizationContext : JSSynchronizationContext
+{
+    private readonly JSThreadSafeFunction _tsfn;
+
+    public JSTsfnSynchronizationContext()
     {
         _tsfn = new JSThreadSafeFunction(
             maxQueueSize: 0,
@@ -26,11 +58,11 @@ public sealed class JSSynchronizationContext : SynchronizationContext, IDisposab
         _tsfn.Unref();
     }
 
-    public void Dispose()
+    public override void Dispose()
     {
         if (IsDisposed) return;
 
-        IsDisposed = true;
+        base.Dispose();
 
         // Destroy TSFN by releasing last thread use count.
         // TSFN is deleted after this point and must not be used.
@@ -67,7 +99,7 @@ public sealed class JSSynchronizationContext : SynchronizationContext, IDisposab
     /// Increment reference count for the main loop async resource.
     /// Non-zero count prevents Node.JS process from exiting.
     /// </summary>
-    public void OpenAsyncScope()
+    public override void OpenAsyncScope()
     {
         _tsfn.Ref();
     }
@@ -76,8 +108,44 @@ public sealed class JSSynchronizationContext : SynchronizationContext, IDisposab
     /// Decrement reference count for the main loop async resource.
     /// Non-zero count prevents Node.JS process from exiting.
     /// </summary>
-    public void CloseAsyncScope()
+    public override void CloseAsyncScope()
     {
         _tsfn.Unref();
+    }
+}
+
+public sealed class JSDispatcherSynchronizationContext : JSSynchronizationContext
+{
+    private readonly JSDispatcherQueue _queue;
+
+    public JSDispatcherSynchronizationContext(JSDispatcherQueue queue)
+    {
+        _queue = queue;
+    }
+
+    public override void Post(SendOrPostCallback callback, object? state)
+    {
+        if (IsDisposed) return;
+
+        _queue.TryEnqueue(() => callback(state));
+    }
+
+    public override void Send(SendOrPostCallback callback, object? state)
+    {
+        if (this == Current)
+        {
+            callback(state);
+            return;
+        }
+
+        if (IsDisposed) return;
+
+        using ManualResetEvent syncEvent = new(false);
+        _queue.TryEnqueue(() =>
+        {
+            callback(state);
+            syncEvent.Set();
+        });
+        syncEvent.WaitOne();
     }
 }
--- a/src/NodeApi/Interop/JSSynchronizationContext.cs
+++ b/src/NodeApi/Interop/JSSynchronizationContext.cs
@@ -3,11 +3,11 @@
 
 using System;
 using System.Threading;
-using Hermes.Example;
+using Microsoft.NodeApi;
 
 namespace Microsoft.JavaScript.NodeApi.Interop;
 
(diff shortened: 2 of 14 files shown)


src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -3,19 +3,51 @@
 
 using System;
 using System.Threading;
+using Hermes.Example;
```
jasongin, line 6: Not used.

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -3,19 +3,51 @@
...
 
 namespace Microsoft.JavaScript.NodeApi.Interop;
 
-public sealed class JSSynchronizationContext : SynchronizationContext, IDisposable
+public class JSSynchronizationContext : SynchronizationContext, IDisposable
```
jasongin, line 10: `abstract` ?

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -3,19 +3,51 @@
...
+
+    public virtual void Dispose()
+    {
+        if (IsDisposed) return;
+        IsDisposed = true;
+    }
+
+    public virtual void OpenAsyncScope() { }
+
+    public virtual void CloseAsyncScope() { }
```
jasongin, line 43: `abstract` ?

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -76,8 +108,44 @@ public void OpenAsyncScope()
...
+
+        if (IsDisposed) return;
+
+        using ManualResetEvent syncEvent = new(false);
+        _queue.TryEnqueue(() =>
+        {
+            callback(state);
+            syncEvent.Set();
+        });
+        syncEvent.WaitOne();
```
jasongin, line 149: If `TryEnqueue()` returned `false`, then it shouldn't wait.

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+        => _getDeferral = getDeferral;
+
+    public JSDispatcherQueueDeferral GetDeferral() => _getDeferral();
+}
+
+public sealed class JSDispatcherQueue
+{
+    private readonly object _queueMutex = new();
+    private List<Action?> _writerQueue = new(); // Queue to add new items
+    private List<Action?> _readerQueue = new(); // Queue to read items from
```
jasongin, line 27: Can the queues' element type be non-nullable `Action`? I don't see any reason it needs to be nullable, except where line 73 sets an index to null after invoking the action, but that seems unnecessary since the queue will be cleared afterward.

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+public delegate void JSTypedEventHandler<TSender, TResult>(TSender sender, TResult args);
+
+public sealed class DispatcherQueueShutdownStartingEventArgs : EventArgs
+{
+    private Func<JSDispatcherQueueDeferral> _getDeferral;
+
+    internal DispatcherQueueShutdownStartingEventArgs(Func<JSDispatcherQueueDeferral> getDeferral)
+        => _getDeferral = getDeferral;
+
+    public JSDispatcherQueueDeferral GetDeferral() => _getDeferral();
```
jasongin, line 20: Can it just return `IDisposable` while keeping the deferral type internal?

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Hermes.Example;
+
+public delegate void JSTypedEventHandler<TSender, TResult>(TSender sender, TResult args);
```
jasongin, line 11: I'm not opposed to typed event-handler delegates in general, but is there any particular benefit of having it here, over normal `EventHandler<TArgs>`?

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+    {
+        lock (_queueMutex)
+        {
+            _deferralCount++;
+        }
+
+        return new JSDispatcherQueueDeferral(() =>
+        {
+            // Decrement deferral count upon deferral completion.
+            TryEnqueue(() => _deferralCount--);
```
jasongin, line 120: Decrementing the deferral count isn't thread-safe. Use a lock, or `Interlocked` methods.

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+                s_currentQueue._threadId = default;
+            }
+
+            s_currentQueue = _previousCurrentQueue;
+        }
+    }
+}
+
+
+public class JSDispatcherQueueController
```
jasongin, line 173: Can these 2 methosd be moved into `JSDispatcherQueue`? I don't see why it's a separate class.

src/NodeApi/JSDispatcherQueue.cs
```diff
@@ -0,0 +1,221 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Hermes.Example;
```
jasongin, line 9: Wrong namespace.

examples/hermes-engine/Program.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Hermes.Example;
+using Microsoft.JavaScript.NodeApi;
+using static Microsoft.JavaScript.NodeApi.JSNativeApi.Interop;
+
+JSDispatcherQueueController controller = JSDispatcherQueueController.CreateOnDedicatedThread();
+controller.DispatcherQueue.TryEnqueue(() =>
```
jasongin, line 9: I'm trying to understand how the dispatcher queue can work, when only Node-API callbacks (not JS callbacks) will use it. For example, consider the following:

```C#
JSValue scriptResult = JSNativeApi.RunScript('new Promise((resolve) => setTimeout(resolve, 1000))');
await ((JSPromise)scriptResult).AsTask();
```

How would the `setTimeout` callback get scheduled (so the Promise gets completed), when the inner callback doesn't get scheduled on the dispatcher queue?

jasongin (changes requested):
Waiting for resolution of discussion about how the dispatcher queue works.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: It is thread safe because we always do it from the `JSDispatcherQueue` thread. I will add a comment.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: It is a separation between a queue and its owner. The queue can be passed around, but only the owner has the right to shutdown it. Splitting up the responsibility/scope is a good thing. E.g. compare `CancellationToken` vs `CancellationTokenSource`.

examples/hermes-engine/Program.cs
vmoroz: The `setTimeout` is not part of the JS engine. It is a part of the hosting environment. For it to work we must add more code to the host. In this case we will need to:
1. Extend the `JSDispatcherQueue` to support timers. It is part of WinRT `DispatcherQueue` - I just did not implement it for now. The implementation typically uses a different `Monitor.Wait` overload with a specified timeout to wake up.
2. Add native implementation for the `setTimeout` polyfill which will create the new timer in the queue in response to this request. 

The way it will work:
- This C# must be exeuted in the `JSDispatcherQueue` task.
- As a part of script execution, the lambda provided to the `Promise` constructor is executed synschonously as a part of the `Promise` constructor.
- The lambda calls the `setTimeout` polyfill function which has C# impleemnt that will schedule a timer job in the queue.
- When the timeout is done it will complete the JS `Promise` which in turn will fulfill the C# `JSPromise` which will continue the execution.

Note that the `JSDispatcherQueue` and the `SynchronizationContext` are the same queue in this scenario.

I should implement support for the async scenarios as the next step to produce more intersting example.

jasongin: OK. That clarifies it for me, thanks. `setTimeout` was just an example, but I guess anything that could possibly schedule future work must be implemented by the host.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: I take it back - you are right - I must use lock here.

src/NodeApi/Interop/JSSynchronizationContext.cs
vmoroz: Fixed.

src/NodeApi/Interop/JSSynchronizationContext.cs
vmoroz: Changed.

src/NodeApi/Interop/JSSynchronizationContext.cs
vmoroz: Changed.

src/NodeApi/Interop/JSSynchronizationContext.cs
vmoroz: Good catch! Fixed.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: Good point! Changed. I use similar approach in my C++ to free captured task resources as soon as possible. I guess it is not a big deal in C# code because the resources cannot be freed immediately anyway.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: Agree. It makes the public surface smaller. Changed.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: Changed. I copied it from the WinRT `DispatcherQueue`, but does not make much sense in this context.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: Fixed.

src/NodeApi/JSDispatcherQueue.cs
vmoroz: Fixed.

examples/hermes-engine/Program.cs
vmoroz: Yes, you are right. I want to add support for timer in the next PR. This PR is already too big and, in some sense, has one piece of complete functionality.

jasongin (approved):

Issue #90, opened 2023-04-12
JSImportAttribute for easier invocation of JS APIs from C#


jasongin:
When `[JSImport]` is applied to a C# `static partial` method, the generated code will load a reference to the JS function (named in an attribute parameter) and marshal the parameters and return value.

Importing entire JS classes could also be supported, with more effort: When `[JSImport]` is applied to a C# `partial class`, the generated code will include a constructor that constructs an instance of the JS type (named in an attribute parameter), and implementations of all methods to marshal parameters and return values.

Issue #59, opened 2023-03-29
Marshalling: Events


jasongin:
.NET events should be projected to JS, somehow.

There is no universal standard API or pattern for JS events; there are slightly different approaches taken for DOM events, Node.js, VS Code, and other projects. (I personally prefer the VS Code events model.)

luanmm:
Personally, I like a lot the Deno's general concept of trying to preserve web standards - a thing that Node broadly neglected til more recently, but seems to be "adjusting its route" to it.

So, just giving an opinion (just theoric) here, I would try to find the most standard way possible in JS that adheres to the .NET standards (like going with delegates or something that could mimic technically the JS behavior).

For reference:
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
- https://nodejs.org/api/events.html#nodejs-eventtarget-vs-dom-eventtarget

Thanks for all the great work, @jasongin!

DevScholar:
Hi, I created a project called [Node PS1 for .NET](https://github.com/DevScholar/node-ps1-dotnet) inspired by the Node API for .NET. It communicates with the .NET Framework via PowerShell 5.1. I used AI-recommended IL-inspired event syntax `add_`/`remove_`.

For example, a WinForms program:
```ts
// Run: node ../node-ps1-dotnet/start.js src/winforms/counter/counter.ts
import dotnet from '@devscholar/node-ps1-dotnet';

dotnet.load('System.Windows.Forms');
dotnet.load('System.Drawing');

const System = dotnet.System as any;
const Forms = System.Windows.Forms;
const Drawing = System.Drawing;

let clickCount = 0;

console.log("--- WinForms Counter (XP Style) ---");

Forms.Application.EnableVisualStyles();
Forms.Application.SetCompatibleTextRenderingDefault(false);

const form = new Forms.Form();
form.Text = "Counter App";
form.Width = 640;
form.Height = 480;
form.StartPosition = 1;

const label = new Forms.Label();
label.Text = "Clicks: 0";
label.Font = new Drawing.Font("Arial", 24);
label.AutoSize = true;
label.Location = new Drawing.Point(90, 30);
form.Controls.Add(label);

const button = new Forms.Button();
button.Text = "Click to Add";
button.Font = new Drawing.Font("Arial", 14);
button.AutoSize = true;
button.Location = new Drawing.Point(100, 90);

button.add_Click(() => {
    clickCount++;
    const message = `Clicked ${clickCount} times`;
[...]

PR #91, opened 2023-04-12
Node.js embedding APIs and demo app

jasongin:
This change adds Node.js embedding APIs using libnode. And it includes a demo app that implements a C# Win UI collaborative text editor using the JS Fluid Framework.

Node.js embedding and other Node API .NET library changes:
 - Add low-level bindings to libnode, using ABI-stable APIs [still in PR].(https://github.com/nodejs/node/pull/43542)
 - For now, the application must provide the path to `libnode`. As discussed, we can consider building and publishing the binaries in some way, until the PR is merged and official binaries are published.
 - Add `NodejsPlatform` and `NodejsEnvironment` classes.
    - The latter starts a dedicated thread when constructed and ends the thread when disposed. I think this functionality belongs in the library because I found it very difficult to get everything in the right sequence for startup and shutdown, so I wouldn't want applications to deal with that challenge.
    - Include an API for enabling inspection for the embedded Node.js process. (It works; I could connect with the Edge debugger.)
 - Add more convenience methods to `JSSynchronizationContext` for running code on the JS thread. (The demo app must often switch back and forth between UI and JS threads.)
 - Fix some minor marshalling bugs.

Demo app:
 - See the readme for an overview. The code is pretty rough now, but it mostly works.
    - There is a known issue that can cause editing to get out of sync when there are concurrent edits from multiple participants. It can be fixed with some more code to keep track of changes between the UI thread and JS thread.
 - The interop code could be significantly cleaned up with some .NET/JS marshalling improvements, particularly [`[JSImport]`](https://github.com/microsoft/node-api-dotnet/issues/90) and [events](https://github.com/microsoft/node-api-dotnet/issues/59).
 - Eventually it would be nice to be able to generate C# interfaces from TypeScript type definitions. The interfaces under the `Fluid` directory in the project are hand-written for now.

https://user-images.githubusercontent.com/13093042/231405696-38bc6831-af13-4b5b-9db8-b29c8e5f5b8c.mp4


src/NodeApi/Engines/NodejsPlatform.cs
```diff
@@ -0,0 +1,89 @@
...
+
+/// <summary>
+/// Manages a Node.js platform instance, provided by `libnode`.
+/// </summary>
+/// <remarks>
+/// Only one Node.js platform instance can be created per process. Once the platform isdisposed,
+/// another platform instance cannot be re-initialized. One or more <see cref="NodejsEnvironment" />
+/// instances may be created using the platform.
+/// </remarks>
+public sealed class NodejsPlatform : IDisposable
```
vmoroz, line 21: >Nodejs [](http://example.com/codeflow?start=20&length=6)

I wonder if we should use use term 'NodeJS' or `Node` instead. I just did not see `nodejs` as one word used often.

vmoroz: Hmm... I just checked Node.js docs and they refer everywhere to Node.js term. You are right. Please ignore my comment.

vmoroz (approved):

src/NodeApi/Engines/NodejsPlatform.cs
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using static Microsoft.JavaScript.NodeApi.JSNativeApi.Interop;
+#if !NET7_0_OR_GREATER
+using NativeLibrary = Microsoft.JavaScript.NodeApi.NativeLibrary;
+#endif
+
+namespace Microsoft.JavaScript.NodeApi.Engines;
```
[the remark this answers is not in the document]
jasongin: Renamed "Engines" to "Runtimes".

jasongin:
@vmoroz I'm going to complete this PR now, but feel free to provide more feedback whenever you have time to review. I have some ideas and plans of my own for improving both the Node.js embedding APIs and the demo app.

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -42,6 +43,148 @@ public virtual void Dispose()
     public abstract void OpenAsyncScope();
 
     public abstract void CloseAsyncScope();
+
+    public void Post(Action action, bool allowSync = false)
```
vmoroz, line 55: >Post [](http://example.com/codeflow?start=16&length=4)

I am finding the function that behaves quite differently depending on the bool parameter to be much more difficult to understand and work with. It is better to keep unique name for a function that can run code synchronously. E.g. we use term InvokeElsePost in Office code.

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -42,6 +43,148 @@ public virtual void Dispose()
...
+    public Task RunAsync(Func<Task> asyncAction)
+    {
+        if (Current == this)
+        {
+            return asyncAction();
+        }
+        else
+        {
+            TaskCompletionSource<bool> completion = new();
+            Send(async (_) =>
```
vmoroz, line 180: >Send [](http://example.com/codeflow?start=12&length=4)

It is better to avoid synchronous call for async invocation.

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -42,6 +43,148 @@ public virtual void Dispose()
...
+        {
+            Post((_) =>
+            {
+                if (IsDisposed) return;
+                _ = asyncAction();
+            }, null);
+        }
+    }
+
+    public void Run(Action action)
```
vmoroz, line 102: >Run [](http://example.com/codeflow?start=16&length=3)

I am trying to understand the need for the new set of Run* functions. Why do we need them in the base JSSynchronizationContext?

src/NodeApi/Engines/NodejsEnvironment.cs
```diff
@@ -0,0 +1,205 @@
...
+    private readonly TaskCompletionSource<bool> _completion = new();
+
+    public static explicit operator napi_env(NodejsEnvironment environment) =>
+        (napi_env)environment._scope;
+    public static implicit operator JSValueScope(NodejsEnvironment environment) =>
+        environment._scope;
+
+    internal NodejsEnvironment(NodejsPlatform platform, string? mainScript)
+    {
+        JSValueScope scope = null!;
```
vmoroz: >JSValueScope scope = null!; [](http://example.com/codeflow?start=8&length=27)

Ideally we must not use the value scope like this. It must be specific to the JS thread.

vmoroz: Overall, we better to keep exactly the same initialization sequence regardless of the runtime we use.

src/NodeApi/Engines/NodejsEnvironment.cs
```diff
@@ -0,0 +1,205 @@
...
+        JSSynchronizationContext syncContext = null!;
+        using ManualResetEvent loadedEvent = new(false);
+
+        _thread = new(() =>
```
vmoroz: >_thread [](http://example.com/codeflow?start=8&length=7)

For the hermes example I use the JSDispatcherQueueController to do that. I wonder if we can reuse it here.

jasongin: I don't see how that could work. There is no task queue here - the queue is hidden inside Node.js.

src/NodeApi/Interop/JSSynchronizationContext.cs
jasongin: The demo app frequently switches between the UI thread and JS thread. I found these patterns were recurring so having them in the library simplifies the application code.

1. The `object state` parameter for the existing `Send()` and `Post()` callbacks is never used in practice, since it is always more convenient (and doesn't require casting) to capture variables with a lambda.
2. It is often necessary to `await` an operation that runs on the JS thread. The `RunAsync()` methods set up a `TaskCompletionSource` for that.
3. Some overloads allow returning a result to the calling thread, either synchronously from `Run<T>()` or asynchronously from `RunAsync<T>()`.

If you prefer, these methods could be moved to extension methods instead of being directly implemented by `JSSynchronizationContext`.

src/NodeApi/Interop/JSSynchronizationContext.cs
jasongin: Yes, this should be `Post`. Also in the other `RunAsync` overload below.

src/NodeApi/Engines/NodejsEnvironment.cs
jasongin: The `JSValueScope` is constructed from within the JS thread. It is only captured and assigned to `_scope` outside the lambda to satisfy the compiler because `_scope` is `_readonly` and therefore must be assigned from the constructor thread.

I don't understand what you mean by "keep the same initialization sequence regardless of the runtime". Other runtimes have different requirements for initialization and threading.

src/NodeApi/Interop/JSSynchronizationContext.cs
jasongin: This method was moved with same semantics from `JSRuntimeContext`. But we can change it.

PR #94, opened 2023-04-13
Improve dynamic JS import for embedding

jasongin:
This change adds some APIs for dynamically importing and converting JS types in .NET, which are then used to somewhat clean up the demo app code.

 - Add `[JSImport]` attribute. It is only used in one small way by the marshaller now (when generating a struct conversion expression) but is likely to be extended later with source-generator support.
 - Expose the `Import()` method on `NodejsEnvironment` class. It was already available on `JSRuntimeContext`, but the former makes more sense for embedding scenarios.
 - Add a generic `Import<T>()` which converts the imported JS value to a .NET type. It is an extension method in the `NodeApi.DotNetHost` assembly because that's where the dynamic marshalling functionality is.
 - Simplify initialization of `NodejsEnvironment`. (Don't create a redundant `JSSynchronizationContext`.)
 - Convert more of the code in the demo app to use .NET types (imported structs and interfaces) instead of `JSValue`.
 - Add some error handling to the demo app around initializing and connecting fluid.
 - Fix a bug in the demo app in which remote carets would not be in the right place on initial join.
 - Update related code snippets in README.

Note there is no source-generation yet for the Node.js embedding scenario. That will be possible later (and could improve startup time), but for now it is only using runtime dynamic bindings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,11 @@ Console.WriteLine('Hello from .NET!');
 ### Minimal example - .NET calling JS
 ```C#
 // C#
-[JSImport("global", "console")]
 interface IConsole { void Log(string message); }
 
 var nodejs = new NodejsPlatform(libnodePath).CreateEnvironment();
 nodejs.Run(() => {
-    var console = nodejs.Import<IConsole>();
+    var console = nodejs.Import<IConsole>("global", "console");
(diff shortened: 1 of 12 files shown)


vmoroz (approved):
:shipit:

Issue #93, opened 2023-04-12
Host throws exception when initialized multiple times by JS workers


jasongin:
The `node-api-dotnet` native host [throws an exception](https://github.com/microsoft/node-api-dotnet/blob/38dc16d596319dce12b10dba0fb10dd4754b89d8/src/NodeApi/DotNetHost/NativeHost.cs#L86) if its `initialize()` function is called multiple times in the same process:

> .NET is already initialized in the current process. Initializing multiple .NET versions is not supported.

Normally a JS `require()` will return the same object if called more than once for the same module path. But multiple initializations can happen for either of two reasons:
  1. Multiple .NET TFMs are requested in the same process, e.g. `require('node-api-dotnet/net6.0')` and `require('node-api-dotnet/net7.0')`.
  2. The module is required (with the same path) from multiple JS workers. Each worker initializes the module separately.

The first case is a valid error, since, as the error message says, multiple .NET versions in the same process are not supported.

But the second case should work. The native host can check if the same .NET version is requested and if so the re-initialization can be allowed. Then the managed host should use a separate assembly load context for each worker.

rijulg:
Just encountered this while calling `load` just once in a github action.

I suspect the github environment has an already runing .Net session which doesn't even let us initalize this module once.

jasongin:
> My dependency graph was actually calling the load function more than once

That should be fine. Subsequent calls to load the same assembly file will just immediately return the already-loaded assembly object.

After looking into this further, I'm not sure my analysis above of the root cause was correct. We actually have a test case that loads .NET from multiple workers in a process, and it works. I'm unable to reproduce the error (except using private APIs that you wouldn't be calling).
 
Can you share:
 - The version of `node-api-dotnet` package you're using
 - Portion(s) of your JS code that use that package - note I do not have access to Office org private repos.

rijulg:
So I have replicated my issue in a public repo (https://github.com/rijulg/node-api-dotnet-issue), unfortunately even the simple example is a bit involved as the bug I encountered was in a project using nestjs so bear with me. I have enapsulated all my project requirements in a devcontainer for running locally, so it should be easy to run locally if you need to.

Essentially we see the following in github actions (but not when running locally):
1. Jest runs the `src/repositories/DLL.spec.ts` test which loads the dll
2. Jest runs the `src/controllers/Example.spec.ts` test which has a dependency on the repositories
[...]

PR #95, opened 2023-04-14
Allow host module re-initialization

jasongin:
Fixes: #93

To resolve the issue, the `NativeHost` instance keeps track of the .NET TFM and managed host path that were used to initialize it, along with a strong reference to the `exports` object that was the result of the initialization. If the initialize function is called again with the same parameters then the same `exports` result is returned.


rijulg (approved):
looks good, waiting to test it out after merge.

PR #96, opened 2023-04-14
Fix NRE in native host JSReference

jasongin:
Fixes: https://github.com/microsoft/node-api-dotnet/issues/93

The `NativeHost` runs in a special "no-context" environment in which `JSRuntimeContext.Current` is null. `JSReference` can support that by getting the `napi_env` handle from the current `JSValueScope` instead of the runtime context.

### DIFF
--- a/src/NodeApi/JSReference.cs
+++ b/src/NodeApi/JSReference.cs
@@ -23,7 +23,7 @@ namespace Microsoft.JavaScript.NodeApi;
 /// </remarks>
 public class JSReference : IDisposable
 {
-    private readonly JSRuntimeContext _context;
+    private readonly JSRuntimeContext? _context;
     private readonly napi_ref _handle;
 
     public bool IsWeak { get; private set; }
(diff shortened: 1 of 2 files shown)


rijulg (approved):
Looks good to me, I tried to solve it and open a PR but you solved it first 😅

src/NodeApi/JSReference.cs
```diff
@@ -118,9 +118,19 @@ protected virtual void Dispose(bool disposing)
...
             napi_ref handle = _handle; // To capture in lambda
-            _context.SynchronizationContext.Post(
-                () => napi_delete_reference((napi_env)_context, handle).ThrowIfFailed(),
-                allowSync: true);
+
+            // The context may be null if the reference was creatd from a "no-context" scope such
+            // as the native host. In that case the reference must be disposed from the JS thread.
+            if (_context == null)
+            {
+                napi_delete_reference((napi_env)JSValueScope.Current, handle).ThrowIfFailed();
```
vmoroz, line 126: We should verify that we are in the JS thread at this point. Could you add a TODO?

jasongin: There is already a check: `JSValueScope.Current` will throw a "no current scope" exception.

PR #98, opened 2023-04-14
Use npm package scope

jasongin:
Change `node-api-dotnet[-generator]` package name to `@microsoft/node-api-dotnet[-generator]` everywhere.


vmoroz (approved):
LGTM

Issue #58, opened 2023-03-29
Evaluate conversions between C# null and JS null/undefined


jasongin:
Consider how C# `null` is mapped to JS `null`/`undefined` in various scenarios, including:
 - C# method that may return a value (reference or value-type) or null
 - C# object with nullable properties (both nullable reference types and `Nullable<T>` value types)
 - JS object with optional properties
 - C# method with optional parameters (and default values which may be `null` or other constants)
 - JS function with optional parameters (as with a callback delegate or an [explicitly imported JS function](https://github.com/users/jasongin/projects/1/views/1?pane=issue&itemId=19819264))

We need to define consistent behavior for how/when C# `null` maps to JS `null` vs `undefined`, for APIs that are automatically marshalled. One rule may not work for all scenarios, so as a backup we may support some kind of marshal attribute to provide a hint to the marshal code generator.

Note the lower-level JS APIs distinguish between `JSValue.Null` and `JSValue.Undefined`, so it will be possible to write C# code that references those explicitly. This issue is about the automatic marshalling of C# APIs that don't directly reference JS types.

vmoroz:
IMHO, the JS `undefined` is mapped much better to C# `null` for the `Nullabe<T>`.
The main reason is that `null` in JS has an `object` type (`typeof`), while `undefined` is always `undefined` that means that there is no value. The JS `null` is unlike C# `null` because it is a special object type. While C# `null` could mean that an optional value type has no value (i.e. undefined).

PR #109, opened 2023-04-22
Marshal null as undefined

jasongin:
Fixes: #58

 - Write doc analyzing null/undefined marshalling considerations
 - Change `JSValue` conversions to convert to `undefined` instead of `null`
 - Change `JSMarshaller` to use `default(JSValue)` instead of `JSValue.Null` or `JSValue.Undefined`
 - Change TS generator to use `?:` or `| undefined` instead of `| null`.
 - Fix test cases that were checking for strict `null` equality.


vmoroz (approved):
:shipit:

Issue #54, opened 2023-03-29
Improve error/exception propagation


jasongin:
So far we have basic exception propagation, but some details are lost, including inner exception messages, and full stack traces.

PR #110, opened 2023-04-22
Merge .NET+JS stack traces

jasongin:
Fixes: #54

It's not perfect: some JS stack frames may be duplicated in the case of a JS->.NET->JS call path. But that will be difficult to fix, and it should not hinder any diagnostics.


vmoroz (approved):
:shipit:

PR #115, opened 2023-04-25
Automate dependency resolution with MSBuild

jasongin:
This greatly simplifies the build and runtime assembly resolution process for the `SemanticKernel` example and similar projects. It leverages MSBuild to copy all dependency assemblies to the output directory and generate type definitions for all of them.

 - TS generator:
    - Enable generating multiple `.d.ts` files with a single command.
    - Omit unused imports from the top of generated `.d.ts` files.
    - Fix handling of non-nullable optional parameters.
    - Don't crash on nested generic types.
    - Add `eslint` suppression comment.
    - Add target assembly name and version comment.
    - Add `--nowarn` option (because you don't want to see >3000 warnings across all assemblies in the SK dependency tree).
  - MSBuild:
    - Suppress output of the primary `.dll` when no code is compiled (`'$(Compile)' == ''`).
    - Copy all package reference assemblies to the output directory, except for the `NodeApi.Generator` assembly and its dependencies.
    - Invoke the TS generator tool on all referenced assemblies to produce `.d.ts` files in the output directory.
 - Simplify SK example loading code (and README).


src/NodeApi.DotNetHost/ManagedHost.cs
```diff
@@ -350,4 +355,20 @@ public JSValue LoadAssembly(JSCallbackArgs args)
         Trace("< ManagedHost.LoadAssembly() => newly loaded");
         return assemblyValue;
     }
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+#if NETFRAMEWORK
+            AppDomain.CurrentDomain.AssemblyResolve += OnResolvingAssembly;
```
vmoroz, line 364: >+= [](http://example.com/codeflow?start=52&length=2)

It should unsubscribe here.

vmoroz (approved):
:shipit:

src/NodeApi.DotNetHost/ManagedHost.cs
jasongin: Fixed.

Issue #108, opened 2023-04-21
Performance benchmarks


jasongin:

jasongin:
[We have some micro-benchmarks now](https://github.com/microsoft/node-api-dotnet/tree/main/bench), which are useful for evaluating the immediate performance impact of code changes.

Eventually I would like to run them automatically for PRs and as part of CI, and track and publish benchmark results over time so we can see trends. But that's not urgent now.

PR #118, opened 2023-04-26
Set up benchmarks project

jasongin:
So far there is just one micro-benchmark: a JS callback. It runs on both AOT and non-AOT (net7.0) runtime.

BenchmarkDotNet does not like its MSBuild output directories modified, so I had to prevent it from using the project-wide `Directory.Build.props`. Fine. It can play in its own directory.

`TestUtils.cs` has some utility functions that are shared by both tests and benchmarks. The source file is included in both projects (there is no project reference).

I think it will be easiest to drive most (or all?) benchmarks from .NET, calling into an embedded Node.js (or other) runtime. And then we will need `libnode` packages published so others (or build machines) can run the benchmarks.

There is a lot more that can be done, but this is a start.

Related to #108

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -36,6 +36,25 @@
                 "reveal": "always"
             },
             "problemMatcher": "$msCompile"
+          },
+          {
+            "label": "bench",
+            "command": "dotnet",
+            "type": "shell",
+            "args": [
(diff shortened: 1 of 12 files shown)


vmoroz (approved):
:shipit:

Issue #65, opened 2023-03-29
Marshalling: Delegates


jasongin:
.NET delegates should be marshaled to & from JS functions.

jasongin:
Not completed.

Issue #61, opened 2023-03-29
Marshalling: Cancellation tokens


jasongin:
It should be possible for JS code to cancel invocation of a .NET API using a cancellation token from a source created in JS.

And it should be possible for JS code to check whether a token passed from .NET to a JS function is cancelled, and throw a cancellation error/exception in that case. (This is perhaps less important because most JS async APIs do not support cancellation.)

Consider mapping .NET `CancellationToken` to/from Node.js [`AbortSignal`](https://nodejs.org/api/globals.html#class-abortsignal)

PR #119, opened 2023-04-28
Delegates and cancellation

jasongin:
- Marshal JS functions to .NET delegates, so .NET code can invoke the JS callback.
   - `Action<>` and `Func<>` as well as custom delegate types are supported.
 - Marshal .NET `CancellationToken` to/from JS `AbortSignal`
 - Add `JSAbortSignal` struct to main library; it facilitates marshalling of `CancellationTokens`.
 - Update TS generator accordingly.

So far, delegate marshalling is only _one-way_: JS can pass a function to .NET, so .NET can then invoke the callback. The opposite (marshalling .NET delegates to JS functions) will come soon in another PR; I just didn't want to block on it.

Fixes: #61 
Partially fixes: #65


vmoroz (approved):
:shipit:

PR #120, opened 2023-04-29
Ensure use of correct env in JSReference

### LINKED ISSUE #105: Fix JSReference verification of napi_ref
The PR #96 broke the validation of `napi_ref` in `JSReference`.
Currently it is possible to use `napi_ref` even if it does not belong to the correct `napi_env`.
It will cause hard to debug crashes and memory corruptions.
We must fix it to ensure the safety of `napi_ref` pointers.

jasongin:
Fixes #105 

Includes new micro-benchmarks for references. There was no measurable difference before and after the change.

### DIFF
--- a/src/NodeApi/JSReference.cs
+++ b/src/NodeApi/JSReference.cs
@@ -23,8 +23,9 @@ namespace Microsoft.JavaScript.NodeApi;
 /// </remarks>
 public class JSReference : IDisposable
 {
-    private readonly JSRuntimeContext? _context;
     private readonly napi_ref _handle;
+    private readonly napi_env _env;
+    private readonly JSRuntimeContext? _context;
 
     public bool IsWeak { get; private set; }
 
@@ -40,8 +41,9 @@ public class JSReference : IDisposable
 
     public JSReference(napi_ref handle, bool isWeak = false)
     {
-        _context = JSRuntimeContext.Current;
         _handle = handle;
+        _env = (napi_env)JSValueScope.Current;
+        _context = JSRuntimeContext.Current;
         IsWeak = isWeak;
     }
 
@@ -68,7 +70,7 @@ public class JSReference : IDisposable
         ThrowIfDisposed();
         if (!IsWeak)
         {
-            napi_reference_unref((napi_env)JSValueScope.Current, _handle, default).ThrowIfFailed();
+            napi_reference_unref(_env, _handle, default).ThrowIfFailed();
             IsWeak = true;
         }
     }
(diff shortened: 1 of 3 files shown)


src/NodeApi/JSReference.cs
```diff
@@ -23,8 +23,9 @@ namespace Microsoft.JavaScript.NodeApi;
 /// </remarks>
 public class JSReference : IDisposable
 {
-    private readonly JSRuntimeContext? _context;
     private readonly napi_ref _handle;
+    private readonly napi_env _env;
```
vmoroz, line 27: >_env [](http://example.com/codeflow?start=30&length=4)

The new code uses unchecked napi_env. It never verifies that the _env is still valid and not points to a deleted memory. I agree that it could be a bit better than before where we used some random current napi_env, but it is not as safe as always going through the JSRuntimeContext that we had originally. It would be great to restore the use of the JSRuntimeContext. We must not allow JSReference creation without valid JSRuntimeContext.

jasongin: > It never verifies that the _env is still valid

Neither does `JSRuntimeContext`. It checks if `_env` is non-null, but that doesn't guarantee it has not been disposed.
```C#
public static explicit operator napi_env(JSRuntimeContext context) => context?._env ??
    throw new ArgumentNullException(nameof(context));
```

> We must not allow JSReference creation without valid JSRuntimeContext.

`NativeHost` needs to create references but should not establish the module context as that needs to be owned by the `ManagedHost`. That is the purpose of the "no-context" value scope type.

jasongin: Also how is this different from `JSValue` which has a `JSValueScope` field, which again only checks for null when converting to `napi_env`?

vmoroz: Then it does not sound right too. `JSValue` used to ensure the safe use of `napi_value` by checking disposed status of the `JSValueScope`. I will check when it is changed. Here we also need a way to ensure that `napi_env` is safe when we use it from the `JSReference`. The initial idea was simple: `napi_env` has instance data. This instance data connected to `JSContext` that was later converted to `JSRuntimeContext`. When `napi_env` is deleted it calls the instance data finalizer that disposes the `JSContext`. The `JSReference` checks the `JSContext` disposed status to make decision if the `napi_env` is still alive. Note that `napi_env` also removes all `napi_ref` when deleted. So, the the wrapped `napi_ref` could also become invalid if we are not aware about `napi_env` deletion.

It is OK to merge this PR "as-is". I will see how to fix it all.

vmoroz (approved):

src/NodeApi/JSReference.cs
jasongin: > JSValue used to ensure the safe use of napi_value by checking disposed status of the JSValueScope. I will check when it is changed.

It's possible some of my refactoring broke that without realizing it, sorry. Clearly we need tests to validate those kinds of protections. But I think that will be difficult without being able to mock out the JS runtime. So I hope we can come up with a workable approach for that, as discussed yesterday.

jasongin: I opened #121 "Add tests to validate checks for mis-use of native handles"

PR #122, opened 2023-04-30
Marshal .NET delegates to JS functions

### LINKED ISSUE #65: Marshalling: Delegates

jasongin:
Fixes #65 

This completes the delegates marshalling work that was started in #119.

 - Marshal .NET delegates to JS functions.
 - When marshalling JS functions to .NET delegates, use `JSSynchronizationContext` to invoke the callback on the JS thread.

### DIFF
--- a/src/NodeApi.DotNetHost/JSMarshaller.cs
+++ b/src/NodeApi.DotNetHost/JSMarshaller.cs
@@ -124,6 +124,7 @@ public class JSMarshaller
 
         if (type.IsGenericTypeDefinition &&
             (type == typeof(Task<>) ||
+            type == typeof(CancellationToken) ||
             type == typeof(IEnumerable<>) ||
             type == typeof(IAsyncEnumerable<>) ||
             type == typeof(ICollection<>) ||
@@ -421,12 +422,6 @@ public class JSMarshaller
     {
         if (method is null) throw new ArgumentNullException(nameof(method));
 
-        if (method.DeclaringType!.BaseType == typeof(MulticastDelegate) &&
-            method.Name == nameof(Action.Invoke))
-        {
-            return BuildToJSDelegateMethodExpression(method);
-        }
-
         try
         {
             string name = method.Name;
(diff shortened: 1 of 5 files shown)


jasongin:
I was considering also creating a `JSFunction` struct that is another `JSValue` wrapper, with explicit conversions to/from delegate types (similar to how `JSPromise` converts to/from `Task`). It could also provide JS function properties/methods like `name`, `apply()`, and `bind()`.

Do you think that would be useful?

jasongin:
> explicit conversions to/from delegate types

Conversions to/from delegate require a `JSMarshaller`, which is not part of the main library because it does reflection / codegen.

But I think a `JSFunction` type in the main library would still be useful. I'll add it in a separate PR.

vmoroz:
It can be useful, but we should be very clear about the dependency graph. Would it make sense to have two different types: one on the main library and another related to dynamic code gen?

---
In reply to: [1529137217](https://github.com/microsoft/node-api-dotnet/pull/122#issuecomment-1529137217) [](http://example.com/codeflow?ancestors=1529137217,1529137217)

vmoroz (approved):
:shipit:

jasongin:
> Would it make sense to have two different types: one on the main library and another related to dynamic code gen?

I was thinking just extension methods for conversions (like `JSPromise`/`Task` conversions), and those could be in the other assembly.

vmoroz:
We need to decide on the criteria where the assembly boundary should be. There can be different dimensions such as: basic Node-API code vs the full JS/C# interop, Node.js specific APIs vs common  JS native API, etc. Before we split it up we can start with defining namespaces/folders first.

---
In reply to: [1529140815](https://github.com/microsoft/node-api-dotnet/pull/122#issuecomment-1529140815) [](http://example.com/codeflow?ancestors=1529140815,1529140815)

jasongin:
We have an assembly boundary already that is based mostly on runtime requirements: AOT compatibility vs not. While we could split up further based on application scenarios, it’s not clear there would be enough benefit in that to justify the added complexity.

vmoroz:
> We have an assembly boundary already that is based mostly on runtime requirements: AOT compatibility vs not. While we could split up further based on application scenarios, it’s not clear there would be enough benefit in that to justify the added complexity.

I agree. There is no point to split assemblies any further at this point. We should just draw the logical boundaries with help of namespaces and folder.

PR #123, opened 2023-04-30
Add JSFunction type

jasongin:
No tests yet, because:
 - This is a thin convenience wrapper around existing APIs.
 - It will be much easier to add tests for this code later when we can mock out the runtime.

### DIFF
--- a/src/NodeApi/JSArray.cs
+++ b/src/NodeApi/JSArray.cs
@@ -34,6 +34,15 @@ public readonly partial struct JSArray : IList<JSValue>, IEquatable<JSValue>
     {
     }
 
+    public JSArray(JSValue[] array)
+    {
+        _value = JSValue.CreateArray(array.Length);
+        for (int i = 0; i < array.Length; i++)
+        {
+            _value.SetElement(i, array[i]);
+        }
+    }
+
     /// <inheritdoc/>
     public int Length => _value.GetArrayLength();
 
--- a/src/NodeApi/JSArray.cs
+++ b/src/NodeApi/JSArray.cs
@@ -36,7 +36,9 @@ public readonly partial struct JSArray : IList<JSValue>, IEquatable<JSValue>
 
     public JSArray(JSValue[] array)
     {
-        _value = JSValue.CreateArray(array.Length);
+        // Do not specify the length because it will create a sparse array
+        // that prevents some V8 optimizations.
+        _value = JSValue.CreateArray();
         for (int i = 0; i < array.Length; i++)
         {
             _value.SetElement(i, array[i]);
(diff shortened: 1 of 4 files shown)


src/NodeApi/JSFunction.cs
```diff
@@ -0,0 +1,357 @@
...
+    private JSFunction(JSValue value)
+    {
+        _value = value;
+    }
+
+    /// <summary>
+    /// Creates a new unnamed JS function with a callback and optional callback data.
+    /// </summary>
+    public JSFunction(JSCallback callback, object? callbackData)
+    : this(JSValue.CreateFunction(name: null, callback, callbackData))
```
vmoroz, line 28: > : [](http://example.com/codeflow?start=3&length=2)

nit: please indent

src/NodeApi/JSFunction.cs
```diff
@@ -0,0 +1,357 @@
...
+        else
+        {
+            JSValue[] thisAndArgs = new JSValue[1 + args.Length];
+            thisAndArgs[0] = thisArg;
+            args.CopyTo(thisAndArgs, 1);
+            return (JSFunction)_value.CallMethod("bind", thisAndArgs);
+        }
+    }
+
+    public JSValue Call() => _value.Call(thisArg: default);
```
vmoroz, line 319: >Call() [](http://example.com/codeflow?start=19&length=6)

It has a different semantic from the JSValue.Call.
Should we keep them aligned and have a different name when function is called as standalone?
JavaScript Function object has `apply` and `call` methods. The JSFunction.Apply corresponds to Function.protype.apply. I would expect the JSFunction.Call would do the same in regard to the Function.protype.call.
The new candidate names are StaticCall and CallAsStatic.

src/NodeApi/JSValue.cs
```diff
@@ -194,11 +193,15 @@ public static unsafe JSValue SymbolFor(ReadOnlySpan<byte> utf8Name)
...
+        string? name, JSCallback callback, object? callbackData = null)
     {
+        if (name is null)
+        {
+            return CreateFunction((byte*)null, 0, callback, callbackData);
+        }
+
 #if NETFRAMEWORK
         int byteCount = Encoding.UTF8.GetByteCount(name);
         byte* utf8Name = stackalloc byte[byteCount];
```
vmoroz, line 207: >stackalloc [](http://example.com/codeflow?start=25&length=10)

Not related to this PR. I just noticed it. are we allowed to use stackalloc for the .Net Framework 4.x? Is it just a C# feature instead of the framework?

src/NodeApi/JSArray.cs
```diff
@@ -34,6 +34,15 @@ public JSArray(int length) : this(JSValue.CreateArray(length))
     {
     }
 
+    public JSArray(JSValue[] array)
+    {
+        _value = JSValue.CreateArray(array.Length);
```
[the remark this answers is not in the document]
jasongin: Well that's nonintuitive.
I'll update it.

src/NodeApi/JSFunction.cs
jasongin: Yes, good point.

vmoroz (approved):
:shipit:

Issue #62, opened 2023-03-29
Marshalling: Ref & Out parameters


jasongin:
While JavaScript does not have any equivalent for `ref` or `out` parameters, it might be too limiting if JS can never call .NET APIs with those parameters. So, there can be some workaround.

As prior art, the WebView2 WinRT interface returns and object with properties for each out parameter:
https://learn.microsoft.com/en-us/microsoft-edge/webview2/how-to/winrt-js-conversion#method-out-parameters
See also: https://devblogs.microsoft.com/oldnewthing/20221124-00/?p=107448

PR #125, opened 2023-05-01
Ref and out params

jasongin:
Fixes: #62

This change enables support for `ref` and `out` parameters for method calls in both directions (JS to/from .NET). It covers static source generation, dynamic invocation, and TS generation.

A .NET method following the `Try*` pattern (with `bool` return value and a single `out` parameter)
is transformed to a JavaScript function with a direct return value of either the result (if the
method returned `true`) or `undefined` (if the method returned `false`).

<table>
<tr><th>C#</th><th>TypeScript</th></tr>
<tr><td>

```C#
bool TryParse(string s, out Version result)
```
</td><td>

```JavaScript
function TryParse(s: string): Version | undefined
```
</td></tr>
</table>

Any other .NET method with `ref` and/or `out` parameters is transformed to a JavaScript function
that returns an object. The object has a `result` property that is the actual return value from
the method (if not `void`), and additional properties with names matching each of the `ref` or
`out` parameters. The `out` parameters do not need to be supplied as arguments in the JS call.

<table>
<tr><th>C#</th><th>TypeScript</th></tr>
<tr><td>

```C#
string[] GetAllResults(ref string value, out int count)
```
</td><td>

```JavaScript
function GetAllResults(value: string):
    { value: string, count: int, result: string[] }
```
</td></tr>
</table>


ericsink:
FWIW, I've tried a similar approach in my implementation, and found one problem:  Any method which uses a "fluent" pattern and which also has a ref/out param will end up projected to TS with an altered return type that doesn't work with the fluent pattern anymore.

jasongin:
@ericsink Is there a certain pattern you have encountered that uses ref/out parameters with the fluent pattern? I'd expect that to be very unusual.

Anyway I don't think there's a good alternative. While it could be possible to simulate ref/out semantics by wrapping arguments in an object, I think that would be messy and unintuitive.

vmoroz (approved):
:shipit:

ericsink:
@jasongin Yep, it's unusual.  I've encountered it when trying to deal with libraries that construct user interface items in a declarative fashion, but they need to capture something in a variable so they express interactions between controls.

For example, the usage might look something like this:

```
new StackPanel()
    .AddChild(
        new Label()
            .Capture(out var lab)
            .FontSize(16)
            .Text("")
    )
    .AddChild(
        new Button()
            .Text("Click Me")
            .Action(_ => lab.Text("Clicked"))
    )
)
```

The idea is that the button click action makes a change to the text of the label.

I also agree that there is no "good" option, but I've been thinking that for this particular pattern, the "simulate ref/out semantics by wrapping arguments in an object" might be the least bad option.

Anyway, I'm just tossing the idea here in case it is of interest, while making no assumptions about what tradeoffs are best for your project.

jasongin:
> The idea is that the button click action makes a change to the text of the label.

I don't understand - that example does not require `ref` parameters. The label is an object that is passed by reference anyway.

ericsink:
The button has a method called Action which accepts a lambda which provides the code to execute on a click.  If that lambda wants to do something to the label, it needs to have captured a reference to same.  And since the overall pattern of constructing the layout doesn't put things in local variables, they introduce things like that Capture() method with an out var param.

Obviously, this code could be done another way, but people do write these kinds of APIs.

jasongin:
> they introduce things like that Capture() method with an out var param

Oh I see - I had missed that when scanning the code before. That is a funny workaround for a problem that I have never had. :) Personally I'm not a huge fan of the "fluent" style, so I would have just created the label in a separate statement. And, JS code can easily do that if needed when calling this kind of API.

PR #127, opened 2023-05-03
Fix TS generator issues after recent changes

jasongin:
Fixing a few issues with building and running the SemanticKernel demo after recent changes:
 - Fix TS generator crash when a `ref` or `out` parameter is both generic and nullable.
 - Fix TS generator output for parameters of `Action<>` and `Func<>` delegates. Parameter names were omitted, but they are required in TypeScript.
 - Fix marshalling of nullable `CancellationToken` parameters.

Also this change will help with diagnosing another error that was reported with using SK APIs:
 - Add a type name parameter to the `Unwrap()` API to improve the error message when unwrapping fails.


vmoroz (approved):
:shipit:

Issue #124, opened 2023-05-01
Help with using collections


rijulg:
While trying to use collections I am running across errors like the following:
```
Class not registered for JS export: System.Collections.ObjectModel.ReadOnlyCollection`1[System.String]
Class not registered for JS export: System.Collections.Generic.List`1
Class not registered for JS export: System.Collections.ArrayList
```

I suspect I am doing something wrong, but a bit hard to find out what the correct process is, can you please help out?

Here's a snippet of my code (the full test project is available here: (https://github.com/rijulg/node-api-dotnet-issue)
```cs
using System;
using System.Collections;
using System.Collections.Generic;
using System.Collections.ObjectModel;

public class HelloDll
{

    internal List<String> _items = new List<String>();
    public ReadOnlyCollection<String> items
    {
        get
        {
            return new ReadOnlyCollection<String>(_items);
        }
    }
    
    public void AddItem(string item)
    {
        _items.Add(item);
    }
}
```

```ts
import { Injectable } from "@nestjs/common";
import { ConfigService } from "@nestjs/config";
import { load as loadDLL } from "node-api-dotnet";

@Injectable()
export default class DLLRepository {
  private readonly helloClass: any;
  constructor(configService: ConfigService) {
    const dllPath = configService.getOrThrow("DLL_PATH");
    const dll = loadDLL(dllPath);
    this.helloClass = new dll.HelloDll();
  }
  addItem(item: string) {
    this.helloClass.AddItem(item);
  }
  getItems(): any {
    return this.helloClass.items;
  }
}

```
Calling `getItems` in JS fails.

jasongin:
Hi, thanks for the issue report.

`ReadOnlyCollection<T>` and `IList<T>` are supposed to work in this kind of scenario, although `ArrayList` isn't currently supported. I will investigate and get back to you.

jasongin:
Oh, you need to use the collection interfaces (`IReadOnlyCollection<T>` or `IList<T>`), instead of the classes (`ReadOnlyCollection<T>`, `List<T>`) on public properties consumed by JavaScript.

The classes may be supported later. But it's a good practice anyway to use collection interfaces for public properties.

jasongin:
[...]

PR #128, opened 2023-05-04
feat: add support for collection classes

rijulg:
* allow getting and setting List
* allow getting ReadOnlyCollection

changes:
- Extracted checks for collection as methods (`IsSettableCollectionType` and `IsGettableCollectionType`) to allow adding more conditions when collections should be wrapped
- Created new extension of ListClass i.e. List<>
- added capability to wrap List<> in builders from and to JS

This PR is in relation to https://github.com/microsoft/node-api-dotnet/issues/124
I am still looking at adding support for, and will open more PRs if things go well.
- Getting and Setting Array
- Getting and Setting ArrayList

### DIFF
--- a/src/NodeApi/Interop/JSCollectionExtensions.cs
+++ b/src/NodeApi/Interop/JSCollectionExtensions.cs
@@ -64,6 +64,13 @@ public static class JSCollectionExtensions
             new JSArrayList<T>((JSValue)array, fromJS, toJS);
 
     /// <summary>
+    /// Creates a list adapter for a JS Array object, without copying.
+    /// </summary>
+    public static List<T> AsListClass<T>(this JSArray array, JSValue.To<T> fromJS, JSValue.From<T> toJS)
+        => ((JSValue)array).IsNullOrUndefined() ? null! :
+            new JSArrayListClass<T>((JSValue)array, fromJS, toJS);
+
+    /// <summary>
     /// Creates an enumerable adapter for a JS Set object, without copying.
     /// </summary>
     public static IEnumerable<T> AsEnumerable<T>(this JSSet set, JSValue.To<T> fromJS)
@@ -385,6 +392,44 @@ internal class JSArrayList<T> : JSArrayCollection<T>, IList<T>
     public void Insert(int index, T item) => Value.CallMethod("splice", index, 0, ToJS(item));
 
     public void RemoveAt(int index) => Value.CallMethod("splice", index, 1);
+}
+
+internal class JSArrayListClass<T> : List<T>, IEnumerable<T>, IEquatable<JSValue>
+{
+    protected JSValue.To<T> FromJS { get; }
+    protected JSValue.From<T> ToJS { get; }
+    private readonly JSReference _iterableReference;
+    internal JSArrayListClass(JSValue array, JSValue.To<T> fromJS, JSValue.From<T> toJS)
+    {
+        ToJS = toJS;
+        _iterableReference = new JSReference(array);
+        FromJS = fromJS;
+    }
+
+    public JSValue Value => _iterableReference.GetValue()!.Value;
+    public new int Count => Value.GetArrayLength();
+
+    bool IEquatable<JSValue>.Equals(JSValue other) => Value.Equals(other);
+
+    public new IEnumerator<T> GetEnumerator() => new JSIterableEnumerator<T>(Value, FromJS);
+
+    System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
+        => GetEnumerator();
+
+    public void Dispose()
+    {
+        // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+        Dispose(disposing: true);
+    }
+
+    protected virtual void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            _iterableReference.Dispose();
+        }
+    }
+
 }
 
 internal class JSSetReadOnlyCollection<T> : JSIterableEnumerable<T>, IReadOnlyCollection<T>
--- a/src/NodeApi.DotNetHost/JSMarshaller.cs
+++ b/src/NodeApi.DotNetHost/JSMarshaller.cs
@@ -1831,6 +1831,10 @@ public class JSMarshaller
                 statements = BuildFromJSToStructExpressions(toType, variables, valueParameter);
             }
         }
+        else if (IsSettableCollectionType(toType))
+        {
+            statements = BuildFromJSToCollectionExpressions(toType, variables, valueParameter);
+        }
         else if (toType.IsClass)
         {
             if (toType == typeof(Stream))
@@ -1862,10 +1866,6 @@ public class JSMarshaller
                         toType),
                 };
             }
-        }
-        else if (toType.IsInterface && toType.Namespace == typeof(ICollection<>).Namespace)
-        {
-            statements = BuildFromJSToCollectionExpressions(toType, variables, valueParameter);
         }
         else if (toType.IsInterface)
         {
@@ -2089,6 +2089,23 @@ public class JSMarshaller
             body: Expression.Block(typeof(JSValue), variables, statements),
             name: delegateName,
             parameters: new[] { valueParameter });
+    }
+
+    private static bool IsSettableCollectionType(Type toType) {
+        var isCollectionInterface = (
+            toType.IsInterface &&
+            toType.Namespace == typeof(ICollection<>).Namespace
+        );
+        Type[] exportedGenericClassCollections = {
+           typeof(System.Collections.Generic.List<>),
+        };
+        var isExportedCollectionType = false;
+        if (toType.IsGenericType) {
+            isExportedCollectionType = exportedGenericClassCollections.Contains(
+                toType.GetGenericTypeDefinition()
+            );
+        }
+        return isCollectionInterface || isExportedCollectionType;
     }
 
     private static bool IsGettableCollectionType(Type fromType) {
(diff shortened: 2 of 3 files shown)


rijulg:
@jasongin can you please review this

rijulg:
I have also added Array and ArrayList support here: https://github.com/rijulg/node-api-dotnet/tree/add-array-and-arraylist-support

However the lack of type information as well as the ability of these objects to have heterogenously typed element is extremely annoying. Essentially we need to support marshalling `System.Object` which being the base of everything in dotnet is almost impossible to fully capture. As a result I was only able to support primitive types using reflection in JSValue for reading, and I stopped after adding support for string for writing.

Would be great if we can merge this PR and then also work on merging that branch to bring in at least partial support.

src/NodeApi/Interop/JSCollectionExtensions.cs
```diff
@@ -387,6 +394,44 @@ internal JSArrayList(JSValue array, JSValue.To<T> fromJS, JSValue.From<T> toJS)
     public void RemoveAt(int index) => Value.CallMethod("splice", index, 1);
 }
 
+internal class JSArrayListClass<T> : List<T>, IEnumerable<T>, IEquatable<JSValue>
```
jasongin, line 397: Unfortunately this class cannot function correctly as a proxy to a JS collection. Most of the methods on the `List<T>` class are not virtual, so they cannot be overridden to access items from the JS collection. And using the `new` keyword on properties and methods here does not help because the object will not be referenced via this internal `JSArrayListClass` type.

I think this means `List` type values can only be marshalled in one direction (.NET to JS), the same as what you have done for `ReadOnlyCollection`.

jasongin (changes requested):

src/NodeApi/Interop/JSCollectionExtensions.cs
jasongin: The alternative would be copy all the values in the collection, instead of trying to make a proxy. Would it be surprising if collection _interfaces_ are marshalled by proxy while collection _classes_ are marshalled by value?

rijulg: I don't think it matters from the JS side how things are handled internally, I thought there was a performance tradeoff (as list items can be massive recursively long objects) but apart from that I can't think of any reason to not use the arrays in that case.

jasongin: It is more than just a performance issue. When JS passes a collection to .NET, and .NET code _modifies_ that collection (by adding/removing/updating items), the JS caller would expect to see the effects of the modification. But that will not work as expected when the collection is marshalled by value.

rijulg: wouldn't the fact that the entire array has been modified have the same effect though? and for specific values within the collection, if the collection is rebuilt with new copies would the original observers die?

rijulg: btw, I have tested the functionality (including my implementation of `Array` and `ArrayList`) with the following code and all of this works fine, so at least the getting and setting part of the `List<>` works as expected

```js
const modPath = require.resolve("node-api-dotnet");
console.log(modPath);

const {load: loadDll} = require("node-api-dotnet");
const assert = require("assert");
const { exit } = require("process");

function iterableToArray(iterable) {
    const array = [];
    for (value of iterable) {
        array.push(value);
    }
    return array;
}

function testStaticsAccess(TestClass) {
    const defaultArray = ["a", "b", "c"];
    assert.equal(TestClass.staticString, "");
    assert.deepEqual(TestClass.staticArray, defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticIList), defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticList), defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticICollection), defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticReadOnlyCollection), defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticArrayList), defaultArray);
    assert.deepEqual(iterableToArray(TestClass.staticArrayClass), defaultArray);
    console.log("testStaticsAccess: Pass");
}

function testInstanceAccess(TestClass) {
    const defaultArray = ["a", "b", "c"];
    const instance = new TestClass();
    assert.equal(instance.instanceString, "");
    assert.deepEqual(instance.instanceArray, defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceIList), defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceList), defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceICollection), defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceReadOnlyCollection), defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceArrayList), defaultArray);
    assert.deepEqual(iterableToArray(instance.instanceArrayClass), defaultArray);
    console.log("testInstanceAccess: Pass");
}

function testStaticsUpdate(TestClass) {
    const testString = "Hello World";
    const testArray = ["p", "q", "r", "s"];
    TestClass.staticString = testString;
    assert.equal(TestClass.staticString, testString);
    TestClass.staticArray = testArray;
    assert.deepEqual(TestClass.staticArray, testArray);
    TestClass.staticIList = testArray;
    assert.deepEqual(iterableToArray(TestClass.staticIList), testArray);
    TestClass.staticICollection = testArray;
    assert.deepEqual(iterableToArray(TestClass.staticICollection), testArray);
    TestClass.staticList = testArray;
    assert.deepEqual(iterableToArray(TestClass.staticList), testArray);
    TestClass.staticArrayList = testArray;
    assert.deepEqual(iterableToArray(TestClass.staticArrayList), testArray);
    TestClass.staticArrayClass = testArray;
    assert.deepEqual(iterableToArray(TestClass.staticArrayClass), testArray);
    console.log("testStaticsUpdate: Pass");
}

function testInstanceUpdate(TestClass) {
    const testString = "Hello World";
    const testArray = ["p", "q", "r", "s"];
    const instance = new TestClass();
    instance.instanceString = testString;
    assert.equal(instance.instanceString, testString);
    instance.instanceArray = testArray;
    assert.deepEqual(instance.instanceArray, testArray);
    instance.instanceIList = testArray;
    assert.deepEqual(iterableToArray(instance.instanceIList), testArray);
    instance.instanceICollection = testArray;
    assert.deepEqual(iterableToArray(instance.instanceICollection), testArray);
    instance.instanceList = testArray;
    assert.deepEqual(iterableToArray(instance.instanceList), testArray);
    instance.instanceArrayList = testArray;
    assert.deepEqual(iterableToArray(instance.instanceArrayList), testArray);
    instance.instanceArrayClass = testArray;
    assert.deepEqual(iterableToArray(instance.instanceArrayClass), testArray);
    console.log("testInstanceUpdate: Pass");
}

function main() {
    const dllPath = process.argv[2];
    const dll = loadDll(dllPath);
    testStaticsAccess(dll.Statics);
    testInstanceAccess(dll.Instance);
    testStaticsUpdate(dll.Statics);
    testInstanceUpdate(dll.Instance);
}

main()
```

```cs
using System;
using System.Collections;
using System.Collections.Generic;
using System.Collections.ObjectModel;

public class Statics
{
    public static string staticString { get; set; } = "";
    public static string[] staticArray { get; set; } = new string[] {"a", "b", "c"};
    public static IList<string> staticIList { get; set; } = new List<string>(new string[] { "a", "b", "c" });
    public static ICollection<string> staticICollection { get; set; } = new Collection<string>(new string[] { "a", "b", "c" });
    public static ReadOnlyCollection<string> staticReadOnlyCollection { get; set; } = new ReadOnlyCollection<string>(new string[] { "a", "b", "c" });
    public static List<string> staticList { get; set; } = new List<string>(new string[] { "a", "b", "c" });
    public static Array staticArrayClass { get; set; } = new string[] {"a", "b", "c"};
    public static ArrayList staticArrayList { get; set; } = new ArrayList(new string[] {"a", "b", "c"});
    public static Array staticArrayInt { get; set; } = new int[] {1,2,3};
}

public class Instance
{
    public string instanceString { get; set; } = "";
    public string[] instanceArray { get; set; } = new string[] {"a", "b", "c"};
    public IList<string> instanceIList { get; set; } = new List<string>(new string[] { "a", "b", "c" });
    public ICollection<string> instanceICollection { get; set; } = new Collection<string>(new string[] { "a", "b", "c" });
    public ReadOnlyCollection<string> instanceReadOnlyCollection { get; set; } = new ReadOnlyCollection<string>(new string[] { "a", "b", "c" });
    public List<string> instanceList { get; set; } = new List<string>(new string[] { "a", "b", "c" });
    public Array instanceArrayClass { get; set; } = new string[] {"a", "b", "c"};
    public ArrayList instanceArrayList { get; set; } = new ArrayList(new string[] {"a", "b", "c"});
}
```

rijulg: @jasongin any advice on what we should do here? considering we potentially don't want to move to array implementation and we can not make editable lists readonly as that significantly breaks away from expected behaviour.

jasongin: > so at least the getting and setting part of the List<> works as expected

It appears to work from the perspective of JS testing because the JS array instance can be round-tripped -- but when .NET tries to use that `List` after it was set by JS it will find it empty because the normal `List` methods have not been overridden to call JS. You can observe this problem by adding a C# method to your class that checks the list count or tries to get an item, and calling that separate C# method from JS.

> any advice on what we should do here

First, I do not expect this to be a common problem because public APIs should use the collection interfaces instead of the classes. See: 
 - [Why we don't recommend using List<T> in public APIs](https://learn.microsoft.com/en-us/archive/blogs/kcwalina/why-we-dont-recommend-using-listt-in-public-apis)
 - [CA1002 - Do not expose generic lists](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1002)

In my experience, all popular .NET class libraries follow that guidance.

But I do think there should be some workarounds for cases when you need to work with some .NET libraries that are not as thoughtfully designed. And it can be OK if the workaround is less convenient/elegant/performant.

As far as I can see there are two options:
  1. Marshal all `List<T>` items by value, similar to how arrays are marshalled now. While the concept and implementation is straightforward, it can lead to unexpected behavior: after JS code modifies the list the changes are not seen by .NET unless the entire list is passed back. And it can lead to unexpectedly bad performance, as the entire (potentially large) list may have to be copied just to access one item, every time. Both of those problems can be avoided by writing careful JS code, but it requires the JS developer to be very aware of the marshalling behavior.
  2. Marshal `List<T>` by proxy in the .NET->JS direction, but marshal by value in the JS->.NET direction. This allows for better behavior and performance in some (perhaps most) use cases, though the asymmetry may be even more confusing than the other option.

If we choose the second option, regular .NET arrays should probably behave the same way. (.NET arrays are currently marshalled by value, in both directions, because they have a similar problem in that the `Array` class cannot be extended, moreover there is no way to create a .NET array over external memory. But like `List<T>`, arrays _could_ be marshalled by proxy in the .NET->JS direction only -- because we can create JS `Proxy` instances for arrays.)

@vmoroz Do you have any thoughts here?

jasongin:
I'm closing this PR since it has been stalled for almost a year.

I still think there is value in doing something like this, but we need to carefully plan the correct approach. I opened #242 to keep track of this, and if/when we revisit the issue we can consider re-opening the PR or re-using some of the code/ideas.

Thanks.

Issue #60, opened 2023-03-29
Marshalling: Tuples


jasongin:
.NET Tuples and ValueTuples should be marshaled to/from JS tuples (aka arrays) with TS per-item type definitions.

Also consider handling .NET KeyValuePair<> in a similar way.

PR #129, opened 2023-05-06
Marshal Tuple, ValueTuple, KeyValuePair

jasongin:
Fixes: #60 

I chose to project .NET tuples as JS arrays rather than JS objects for two reasons:
 - [TypeScript uses the term "tuple"](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) to refer to "an array type that knows exactly how many elements it contains and exactly which types it contains at specific positions".
 - JavaScript core APIs such as [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) use array tuples, not objects, for key-value pairs.

For consistency, `ValueTuple`s with named properties are still projected as JS arrays, not JS objects with named properties. So C# `(string A, int B)` becomes TypeScript `[string, number]`, not `{ A: string, B: number }`.

This change also adds support for the `Predicate<T>` generic delegate, which is commonly used in public APIs. I forgot to include along with `Action<>` and `Func<>` delegate support added recently.

### DIFF
--- a/Docs/typescript.md
+++ b/Docs/typescript.md
@@ -26,17 +26,25 @@ generated automatically. Or, the generator tool is available separately via the
 | `string` | `string`              |
 
 ### Object types
-
-| C# Type                       | TypeScript Projection            |
-|-------------------------------|----------------------------------|
-| `class Example`               | `class Example`                  |
-| `struct Example`              | `class Example`                  |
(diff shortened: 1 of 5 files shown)


vmoroz (approved):
:shipit:

PR #130, opened 2023-05-07
Simplify native calls

vmoroz:
In the current code each native call checks if function is loaded and loads it on demand.
It makes each call to do extra check.
In this PR we simplify each call:
- During initialization we assign to each native pointer a special delay load stub.
- The delay load stub loads the function and assigns its pointer to the function pointer field.
- All other calls through the function pointer are going to call directly the native functions.

### DIFF
--- a/src/NodeApi/Native/JSNativeApi.Interop.cs
+++ b/src/NodeApi/Native/JSNativeApi.Interop.cs
@@ -606,6 +606,10 @@ public static partial class JSNativeApi
                 // js_native_api.h APIs
                 //----------------------------------------------------------------------------------
 
+                // Use explicit delegate types to make sure that we use a delegate with CDecl
+                // calling conventions and to avoid runtime error when calling
+                // Marshal.GetFunctionPointerForDelegate because it cannot accept generic
+                // delegate types.
                 napi_add_finalizer = (delegate* unmanaged[Cdecl]<
                     napi_env, napi_value, nint, napi_finalize, nint, nint, napi_status>)
                     GetFunctionPointerForDelegateAndRootIt<DelegateTypes.napi_add_finalizer>(
--- a/src/NodeApi/Native/JSNativeApi.Interop.cs
+++ b/src/NodeApi/Native/JSNativeApi.Interop.cs
@@ -9,10 +9,6 @@ using System.Runtime.InteropServices;
 using System.Security;
 using System.Text;
 
-#if NET7_0_OR_GREATER
-[assembly: DisableRuntimeMarshalling]
-#endif
-
 namespace Microsoft.JavaScript.NodeApi;
 
 public static partial class JSNativeApi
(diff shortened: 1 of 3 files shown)


src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -1478,358 +2796,2267 @@ internal static bool TryGetExport(ref nint field, [CallerMemberName] string func
...
         {
-            nint funcHandle = GetExport(ref field, functionName);
-            var funcDelegate = (delegate* unmanaged[Cdecl]<
-                napi_env, nint, nint, napi_status>)funcHandle;
-            fixed (TResult1* result1_native = &result1)
-            fixed (TResult2* result2_native = &result2)
-            {
-                return funcDelegate(env, (nint)result1_native, (nint)result2_native);
-            }
+            GCHandle.Alloc(d); // To make sure that the delegate is not collected.
```
jasongin, line 2812: I guess this is cleaner than what I did to prevent the "JS invoke" callback delegates from being GC'd, which was to save them in static fields. See `JSNativeApi.s_invokeJS*` - maybe those should use this approach also?

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -39,168 +39,1241 @@ public static void Initialize(nint libraryHandle = default)
...
+                napi_destroy_platform = &DelayLoadStubs.napi_destroy_platform;
+                napi_run_environment = &DelayLoadStubs.napi_run_environment;
+#else
+                //----------------------------------------------------------------------------------
+                // js_native_api.h APIs
+                //----------------------------------------------------------------------------------
+
+                napi_add_finalizer = (delegate* unmanaged[Cdecl]<
+                    napi_env, napi_value, nint, napi_finalize, nint, nint, napi_status>)
+                    GetFunctionPointerForDelegateAndRootIt<DelegateTypes.napi_add_finalizer>(
```
jasongin, line 611: Can the explicit type arg be omitted from all these calls to `GetFunctionPointerForDelegateAndRootIt()`? It looks like it can be implicit.

jasongin (approved):

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: I initially implemented static fields too following your approach, but then I asked ChatGPT about the GetFunctionPointerForDelegate usage and how to keep the delegate alive and it gave code that used the `GCHandle.Alloc`. It is just amazing what it can do if we can ask it right questions. :)

I will see how to improve the code for callbacks in the follow up PR.

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: I does not work. It seems that if I omit the explicit delegate type, then the compiler uses Func<...> specialization and function wilI fail because GetFunctionPointerForDelegate does not like generic delegates.
I will add a comment about it.

Issue #67, opened 2023-03-29
Marshalling: Generics


jasongin:
JavaScript code should be able to:
  - Call generic methods (with explicit or _maybe_ implicit type arguments)
  - Get a property value or return value that is an instance of a .NET generic class, struct, or interface
  - Construct a new instance of a .NET generic class or struct (with explicit type arguments)
  - Get properties or invoke (non-generic) methods on an instance of a generic class, interface, or struct
  - Get static properties or invoke (non-generic) static methods on a generic type
  - Implement a .NET generic interface in a JavaScript class and pass an instance of it to .NET
  - Invoke or implement generic delegates

Note certain special-case generic types with mappings to built-in JavaScript types are implemented separately, including generic collections and generic tasks. This is about the general case.

In C#, type arguments can often be implicit, based on the argument types. Implicit type arguments often will not work well with JS since the type cannot as often be inferred from the JavaScript argument values. So JS code will usually need to specify explicit type arguments.

jasongin:
## Projecting .NET Generics to JavaScript

Since the JavaScript runtime type system lacks generics, the projection of .NET generic APIs into
JavaScript needs to be considered carefully.

Note while TypeScript _does_ have generics, they are merely a compile-time facade and do not have
any effect on runtime binding or execution. So it is unfortunately not possible in most cases
to directly map .NET generics to TypeScript generics. (The exception is when mapping .NET generic
collections to TypeScript generic collections like `Array<T>` and `Map<TKey, TValue>`, which does
work.)

Following is a non-exhaustive list of scenarios for .NET generic APIs that the JavaScript interop
mechanism needs to support:

| Scenario | Example (C# syntax) |
|----------|---------------------|
| Calling a generic method                             | `Enum.Parse<T>()`
| Getting/setting a static property on a generic class | `Comparer<T>.Default`
| Calling a static method on a generic class           | `Comparer<T>.Create()`
| Creating an instance of a generic class              | `new KeyValuePair<TKey, TValue>()`
| Methods with both generic and non-generic overloads  | `Enum.Parse<T>()`<br/>`Enum.Parse()`
| Generic and non-generic classes with same name       | `TaskCompletionSource<T>`<br/>`TaskCompletionSource`

In general, JavaScript code needs a way to specify generic type arguments when working with .NET
generic types and generic methods. There are a few different syntax approaches that might be
considered for specifying type arguments...

### Rejected syntax (A)
[...]

PR #131, opened 2023-05-08
Generic interface method callbacks

jasongin:
This change enables JS to implement .NET interfaces that include _generic methods_, which can then be called by .NET with arbitrary generic type arguments. A primary motivation is to support the [`Microsoft.Extensions.Logging.ILogger` interface](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger.log) that is used by the `SemanticKernel` example.

 - Add a `JSMarshaller.Current` property. It can be used by code that needs to do dynamic marshalling to get the current marshaller instance (with appropriate settings and caches).
    - Unfortunately the current marshaller instance cannot be a property on `JSRuntimeContext` because of assembly separation. (Maybe it should be an extension method on that class?)
    - `ManagedHost`, `JSInterfaceMarshaller`, and `JSRuntimeContextExtensions` are updated to reference this property.
 - Rename a few `JSMarshaller` methods for greater clarity.
 - Add support for generic methods and parameters in `SymbolExtensions` when converting from method symbols and type symbols to methods and types.
 - Make small updates in a few other areas of the marshalling code to support generic type parameters.
 - Add support for generic interface methods in `JSInterfaceMarshaller`
 - Marshal unknown (`object`) or nonpublic types as JS "external" values. They cannot be used from JS, but at least they can be round-tripped back to .NET. (We can consider doing something a little better later.)
 - Get actual exception message from `TargetInvocationException` which may be thrown from dynamic delegate invocations.
 - Add a test case for a generic interface method implemented by JS. It is skipped for AOT testing because AOT cannot support that kind of dynamic marshalling.
 - Update `examples/SemanticKernel` to implement `ILogger`.

There is still a lot more to do to enable [full generics marshalling](#67), but this is a small piece of it.

### DIFF
--- a/examples/semantic-kernel/example.js
+++ b/examples/semantic-kernel/example.js
@@ -1,17 +1,33 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { SK, SKOpenAI } from './semantic-kernel.js';
+// @ts-check
 
-const kernel = SK.Kernel.Builder.Build();
+import { SK, SKOpenAI, Logging } from './semantic-kernel.js';
+const LogLevel = Logging.LogLevel;
+
+/** @type {import('./bin/Microsoft.Extensions.Logging.Abstractions').ILogger} */
+const logger = {
+  Log(logLevel, eventId, state, exception, formatter) {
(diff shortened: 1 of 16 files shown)


vmoroz (approved):
:shipit:

PR #135, opened 2023-05-09
Fix optional-parameters bug

jasongin:
Overload resolution was selecting the wrong index for the default parameter value, which could cause an error similar to:

> Null default value is not valid for parameter type: System.Double

I'm also adding test cases for optional parameters, since there were none before.

Update: Fixing an issue that has been failing PR builds recently. The TS generator could fail if invoked multiple times on the same build-server thread.


vmoroz (approved):
:shipit:

PR #136, opened 2023-05-09
Fix regression when passing value as `object` type

jasongin:
This fixes a regression when working with APIs such as `IKernel.ImportSkill(object)`.

If a .NET property or method takes an `object` type, it should accept _either_ a .NET object wrapped as a JS value or a JS external value. (The latter is used for .NET objects of unknown or non-public types.)

Also check for `IsNestedPublic` when deciding whether to marshal as external. Oddly, nested types have `IsPublic=false` and `IsNestedPublic=true`. There is more work still to fully [support nested types](#133), but this may unblock some scenarios.


vmoroz (approved):
:shipit:

PR #142, opened 2023-06-04
Generics marshalling

jasongin:
Fixes: #67

 - Dynamically export generic types and methods using the `$` suffix convention as described in `generics.md`.
 - Generate type definitions for generic types.
   - Aside from the generic factory function, generic classes are projected as two TS interfaces: `ClassName$$N` and `ClassName$N` (where N is the number of generic type parameters). The `$$` interface includes static members and constructors; the other includes only instance members. (These names will rarely if ever need to be typed out in user code.)
 - Add test cases for generic invocation and typedefs generation.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -38,6 +38,12 @@
 
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning" PrivateAssets="all" />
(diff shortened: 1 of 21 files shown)


vmoroz (approved):
:shipit:

PR #145, opened 2023-06-12
fix: Update type generator to use new syntax of load

rijulg:


rijulg:
@jasongin can you please review this PR?

jasongin (approved):

jasongin:
Thanks. I didn't catch this because all my projects use the MSBuild targets in the nuget package to invoke the generator, so they don't use this script.

PR #153, opened 2023-09-17
Replace .NET 7 target with .NET 8 rc1

### LINKED ISSUE #116: .NET 8 Support

jasongin:
Fixes #116

I am replacing the .NET 7 target with .NET 8. Version 7 is not an LTS release, while 6 and 8 are, so it makes sense to support .NET versions 472, 6, and 8 for the near future.

.NET 8 is now RC1 status which has a go-live license. Merging this change will require .NET 8 RC1 SDK to develop/build the `node-api-dotnet` project, but NOT to develop applications that use it. (Any of .NET 472, 6, 7, or 8 runtimes should still work.) Of course, we'll update from RC1 to the released .NET 8 version when it's available in a couple months.

This update required more work than just replacing `7` with `8` in a few places:
 - The code for loading .NET `HostFxr` is updated to allow a prerelease version (at least for now).
 - .NET 8 disables dynamic code generation when compiling with `PublishAot=true` property. For most applications it makes sense to have the managed code behave the same as the Native AOT compiled app. But `node-api-dotnet` depends on `Reflection.Emit` in its managed host (not in its native host). To work around that I added a special `aot` target, with separate output path, that is built and published when packaging (and for tests).
 - The source generator assemblies are in a "collectible" assembly load context in .NET 7 and 8, which is incompatible with the default behavior of `Expression.Lambda()`. I had previously used [a workaround for .NET 7 (to reload the generator assemblies as non-collectible)](https://github.com/microsoft/node-api-dotnet/commit/a2511f0beb17a60ae82f542411fe961d2f44f6aa) but that doesn't work anymore on .NET 8. This change includes code to explicitly create the lambda delegate types in a collectible context, which resolves the issue. It is a lot more code but is a better solution since now the source-generator can be properly unloaded (and reloaded to pick up changes).
 - The `VSTestNoBuild` behavior changed a bit in the .NET 8 SDK, so I had to update the logic that is used to conditionally skip the dependency builds during the tests.
 - The .NET 8 SDK added more analyzer rules checked by `dotnet format`. So there are small updates in several places to resolve those.

Additionally, I noticed the .NET 6 targeted tests were mistakenly using the .NET 7 or 8 runtime. The build and host logic is updated to fix the issue.


jasongin:
There's an issue with the source-generator on Max & Linux that causes most of the tests to fail. I know a workaround, but it will be somewhat complicated to implement.

PR #157, opened 2023-10-04
Add benchmarks, GC handle tracking test

jasongin:
- Add more micro-benchmarks covering a variety of basic .NET <-> JS calls.
 - Add GC handle tracking: track just a handle count in release mode, and allocation stack in debug mode.
 - Add a test case that uses the handle-tracking to verify proper GC after constructing a .NET instance from JS, then releasing it.
 - Implement string array marshalling for `napi_create_platform` (to support `--enable-gc` option)
 - Add `TryGetValueExternal()` API, which returns `null` if the value is not an external. This fixes a bug in marshalling that arose when constructing a `ManagedHost` instance for benchmarks.

I added GC handle tracking because I was struggling with GC pressure during the many iterations that BenchmarkDotNet likes to run. First I thought there was a leak, but the test confirmed there is not. The problem is just that BenchmarkDotNet can run millions of iterations, which may each allocate and release objects, and JS GC does not keep up, so successive iterations get gradually slower. I couldn't figure out a way to explicitly run the JS GC during tests without distorting the benchmark results. (The test runner carefully manages the .NET GC, but doesn't know anything about JS.) So the workaround for now is to reduce the number of iterations.

### DIFF
--- a/bench/Benchmarks.cs
+++ b/bench/Benchmarks.cs
@@ -7,80 +7,265 @@ using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Running;
+using Microsoft.JavaScript.NodeApi.DotNetHost;
+using Microsoft.JavaScript.NodeApi.Interop;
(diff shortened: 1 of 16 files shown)


vmoroz (approved):
:shipit:

PR #159, opened 2023-10-12
Refactor native APIs

jasongin:
This refactor moves all native APIs (`napi_*` functions) from the `JSNativeApi.Interop` class to a `JSRuntime` abstract class. Then a specific runtime (`NodejsRuntime`) implements the abstract class via (delayed) binding to the native APIs, while also handling some low-level details like GC pinning.

This has a few benefits:
 - API calls can be overridden by tests, either for full or partial mocking of the runtime or for specific behaviors like error injection. I have not yet written new tests that take advantage of this, but I plan to use it to test object lifetime and threading errors.
 - API calls can be traced by injecting a wrapper around the runtime API, with zero cost when tracing is not enabled (the wrapper is not injected). I am already finding this VERY helpful for debugging and performance optimization. See the `TracingJSRuntime` class.
 - It could be possible to multiple runtimes in the same process, since the native API bindings are no longer static.

The drawback is this adds another level of indirection: every native API call is a virtual method invocation (along with a null check for delay-loading). I know there was a lot of work (and code) put into avoiding both of those costs. But now that we have micro-benchmarks, we can fairly accurately measure the impact of the changes... or lack of impact. From what I can see so far, there is **no measurable performance impact**: all of the benchmark results are consistently within the margin of error (~4%) before and after the changes. (Sometimes a bit slower, sometimes a bit faster.) I still need to do some more perf measurements on other OS's.

At this point all tests are passing, but some API areas are not yet refactored. And, I want to think more about how to initialize and reference the current runtime thread-static in relation to the current `JSValueScope` and `JSRuntimeContext` thread-statics.


vmoroz:
I like the overall direction of making the Node-API low implementation pluggable.
My concerns mostly are:
- We should avoid extra memory allocations if we can. The new API replaces all uses of char spans with strings. I would rather see them to co-exist. One is more efficient, while another is more convenient. The char span based APIs can be virtual, while string-based can be convenience wrappers on top of them.
- While using virtual classes and inheritance seems to be the simplest solution, the alternative could be just providing an alternative set of native function pointers. I do not know which one is better. I have seen that Meta's JSI being based on the interface provides very good flexibility in a number of cases.
- In JSI for Node-API where I implemented the same flexible approach, I found that the biggest challenges were:
  - Tracking the current set of APIs across multiple threads. We must be very careful with it and provide the correct context. In some cases I chose eager function binding to avoid issues.
  - Dynamic loading if interconnected APIs and providing default implementation if they miss. Such APIs must be loaded as a group or defaulted as a group. E.g. loading/saving preprocessed byte code. It made me think if it is worth loading native APIs on demand. Maybe it is better to load them all at once and do not waste time on each call. In C++ I managed avoiding the runtime check by substituting the whole function. I wonder if we can do the same here.

src/NodeApi/Runtime/NodeJSRuntime.cs
```diff
@@ -0,0 +1,2277 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using static Microsoft.JavaScript.NodeApi.JSNativeApi.Interop;
+
+namespace Microsoft.JavaScript.NodeApi.Runtime;
+
+public unsafe class NodejsRuntime : JSRuntime
```
vmoroz, line 11: Instead of re-writing the whole low-level API from scratch, it would be better to implement it as a layer on top of the existing JSNativeApi.Interop.cs.
Loading of native functions from a specific DLL must be static anyway. 
The new virtual interface on top of it would allow the required flexibility to either use specific DLL API directly or add tracing on top of it. We can also load it from different DLL.

src/NodeApi/Runtime/NodeJSRuntime.cs
```diff
@@ -0,0 +1,2277 @@
...
+        napi_run_environment;
+
+#pragma warning restore CS0169
+#pragma warning restore IDE0044
+
+    private delegate* unmanaged[Cdecl]<napi_env, nint, napi_status> napi_get_version;
+
+    public override napi_status GetVersion(napi_env env, out uint result)
+    {
+        if (napi_get_version == null)
```
vmoroz, line 203: Checks like this are unnecessary. Our current code has a better implementation based on stubs and replacing them with real function pointers.

src/NodeApi/Runtime/NodeJSRuntime.cs
```diff
@@ -0,0 +1,2277 @@
...
+    public override napi_status ThrowError(napi_env env, string? code, string msg)
+    {
+        if (napi_throw_error == null)
+        {
+            napi_throw_error = (delegate* unmanaged[Cdecl]<napi_env, nint, nint, napi_status>)
+                GetExport(nameof(napi_throw_error));
+        }
+
+#if NET5_0_OR_GREATER
+        Span<byte> codeBytes = stackalloc byte[Encoding.UTF8.GetByteCount(code ?? string.Empty)];
```
vmoroz, line 390: We must be careful with such unbounded stack allocations like this. They can lead to stack corruption and related security risks. It would be much better to do the allocations up to certain limit, and then use heap for larger strings. The Lib import generated code uses a special class to do it.

src/NodeApi/Runtime/NodeJSRuntime.cs
jasongin: Yes, but that required a lot of extra code to make that work, which adds to maintenance overhead. The cost of a simple null check is negligible (as confirmed by benchmarks), so I prefer the simpler code.

jasongin:
> We should avoid extra memory allocations if we can. The new API replaces all uses of char spans with strings.

Not all: just function names, symbol names, and error messages, which are almost always going to already be `string` type in the C# code. The `GetValueString*` APIs still use `Span<T>`. But I would be OK with moving all string encoding back to the layer above this.

> the alternative could be just providing an alternative set of native function pointers

I considered that, but chose the inheritance approach because:
  - Inheritance is easier and more flexible to work with, e.g. to override individual methods.
  - A virtual method call is not any more expensive than an indirect call through a pointer to a set of functions (because that is essentially what a v-table is).
  - It is more convenient to be able to override the functions at a slightly higher level, to work with `Span<>` and `out` parameters rather than raw pointers.

> Maybe it is better to load them all at once and do not waste time on each call.

I was assuming that would impact startup time, but maybe it's not much. I can try to benchmark it.

src/NodeApi/Runtime/NodeJSRuntime.cs
jasongin: Yes, good point. I'll fix it. (This code was refactored from `JSValue.cs`, so the issue existed there already.)

src/NodeApi/Runtime/NodeJSRuntime.cs
jasongin: > Instead of re-writing the whole low-level API from scratch, it would be better to implement it as a layer on top of the existing JSNativeApi.Interop.cs.

I considered it more of a refactor than a rewrite. I don't see a benefit to having another lower layer.

> Loading of native functions from a specific DLL must be static anyway.

Yes, but this enables different threads to bind functions from different DLLs or different versions of a DLL. So an application could host multiple JS engines using multiple versions of `libnode.dll`. (Why would anyone want to do that? IDK, but I always find that avoiding static data allows for more flexibility.)

vmoroz: I wonder what is your approach for the .Net Framework 4.8? One of the reasons we had these static points is to address the delegate lifetime issues.

vmoroz: Regarding loading multiple dlls: one scenario as we have it in RN for Windows is to load V8 and Hermes DLL. It is quite a valid point. In C++ I had to use templates that allowed me to have static pointers and still have them to be different per JS engine. I agree with you that this new approach is much simpler and if we do not have much overhead, then it is great.

vmoroz:
> > We should avoid extra memory allocations if we can. The new API replaces all uses of char spans with strings.
> 
> Not all: just function names, symbol names, and error messages, which are almost always going to already be `string` type in the C# code. The `GetValueString*` APIs still use `Span<T>`. But I would be OK with moving all string encoding back to the layer above this.

I see. If it is only for errors, then it must be fine. For errors we will have much more overhead anyway.
> > the alternative could be just providing an alternative set of native function pointers
> 
> I considered that, but chose the inheritance approach because:
> 
> * Inheritance is easier and more flexible to work with, e.g. to override individual methods.
> * A virtual method call is not any more expensive than an indirect call through a pointer to a set of functions (because that is essentially what a v-table is).
> * It is more convenient to be able to override the functions at a slightly higher level, to work with `Span<>` and `out` parameters rather than raw pointers.

It makes sense. Thanks!

> > Maybe it is better to load them all at once and do not waste time on each call.
> 
> I was assuming that would impact startup time, but maybe it's not much. I can try to benchmark it.

It would be interesting to see. Maybe the whole code could become much simpler.
Though, the code as it is already looks much simpler.

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -4,15 +4,32 @@
...
 using System.Text;
+using Microsoft.JavaScript.NodeApi.Runtime;
 
 namespace Microsoft.JavaScript.NodeApi;
 
 public static partial class JSNativeApi
 {
+    // TODO: When complete, JSRuntime will mostly or fully replace the JSNativeApi class.
+    // TODO: Move this to a thread-local or attach it to the thread-local
+    // JSValueScope or JSRuntimeContext.
```
vmoroz, line 20: My vote is for attaching it to `JSValueScope` and/or `JSRuntimeContext`. It should help associating JS values/objects with the correct runtime regardless of the thread.

src/NodeApi/Runtime/NodeJSRuntime.cs
jasongin: The delegate lifetime issue was when we'd pass an unmanaged delegate to native code without holding onto it in managed code. We dealt with that for callbacks, but it was never an issue with the regular native APIs.

Tests are passing on .NET 4.7.2 with the latest update.

src/NodeApi/Runtime/NodeJSRuntime.cs
jasongin: Fixed, using `ArrayPool` and a new `PooledBuffer` class that makes it convenient to lease the pooled buffers via `using`.

Now even short string are using pooled buffers rather than `stackalloc`. I couldn't work out a way to use either `stackalloc` or pooled (heap) buffers depending on length, without making the code very messy, due to the restriction on passing fixed or `Span<>` values into a lambda.

vmoroz (approved):

jasongin:
I just pushed a big update to this PR:
 - All native APIs are moved into the new `JSRuntime` abstraction. The `JSNativeApi.Interop` class is eliminated.
 - Optional tracing is implemented for all APIs.
   - Function names are tracked in `JSCallbackDescriptor` / `JSPropertyDescriptor` to enable better tracing.
 - A generic `Import()` helper method eliminates the `if (func != null)` checks before every API. The check still happens but it's only in the helper instead of repeated everywhere.
   - I benchmarked the alternative approach of pre-binding all of the APIs. It was a little bit slow: around a millisecond on the first call vs ~80 nanoseconds for subsequent calls. (And there was no measurable difference from removing the `if` check.) Since the motivation would have been mostly just fewer lines of code, I don't think it's worth it after the other `Import()` change achieved most of the same benefit.
 - Since the native API bindings are in a `JSRuntime` instance, rather than static, that instance needs to be maintained with context. I attached it to `JSValueScope` because many API calls already have a scope instance conveniently available. Some other APIs still need to get the runtime via the current (thread-local) scope.
 - I rewrote the `JSValueScope` constructor to accommodate passing in a `JSRuntime` instance, and also to add more consistency checks related to the relationships between different types of scopes.
 - I realized we can use `[CallerArgumentExpression]` by defining the attribute type for .NET 4.x. So it is used again in a few places.
 - The C# language version is updated to `12`. My favorite addition is the simpler (JS-like) syntax for arrays: `[a, b]` instead of `new[] { a, b }`

src/NodeApi/Native/JSNativeApi.Interop.cs
jasongin: Done as (non-static) `JSValueScope.Runtime` and `static JSValueScope.CurrentRuntime => JSValueScope.Current.Runtime`.

PR #180, opened 2023-11-12
Handle scope validation tests

### LINKED ISSUE #121: Add tests to validate checks for mis-use of native handles
Mis-use of native handles can lead to crashes, memory corruption, or other bad behavior. This includes:
 - Accessing any `napi_value` or `napi_ref` from the wrong thread.
 - Accessing any `napi_value` or `napi_ref` after the associated `napi_env` has been disposed.

We should add test cases to ensure those kinds of mistakes are detected and result in reasonably helpful exceptions.

jasongin:
Fixes #121 

 - Add unit tests for handle scope validations, using a mocked runtime so that the tests can run without `libnode`. [1]
   - Test rules for nesting different types of value scopes.
   - Test attempts to access values from disposed scope or from different thread.
   - Test attempts to access references from a different thread.
   - The checks should also cover attempts to access values/references from a disposed environment because they rely on the thread-local current scope that will be null (or different) after the environment root scope is closed. There are no explicit tests for this though.
  - Refactor and fix handle checks in `JSValue`, `JSValueScope`, and `JSReference` to pass the tests.
  - When making API calls related to a `JSValue` instance, always use the value's `Scope` member rather than the thread-local "current" scope. This allows for validation that the current and target environments (threads) are the same.
  - Add new exception types `JSValueScopeClosedException` (subclass of `ObjectDisposedException`) and `JSInvalidScopeException` (subclass of `InvalidOperationException`) to provide detailed error info when a handle scope problem is detected.
  - Pass a GC handle to the `JSRuntimeContext` via the hint parameter of all finalizers (which required some minor refactoring where the hint was used for other things). Previously the finalizers were getting the context via the thread-local scope, which caused errors when finalizers ran after the root scope was closed. Also the finalizers were calling `napi_get_instance_data` (many times) which should be avoided. Finalizers now do not make any JS calls and do not rely on any thread-statics.
  - Fix a similar issue with `JSReference.Dispose()`, to handle disposing references after the root scope was closed.
  - Fix a bug in generator tool response file parsing of an empty quoted value (`""`). This broke builds targeting .NET 4 where the `--packs` argument value is empty. (I don't know how this passed testing before.)

I compared benchmark results before and after the changes - there was no measurable impact from the additional validations.

[1] Currently I get a crash when trying to create more than one "environment" using the new embedding APIs. I will check whether that is fixed in a newer revision of that PR; I'm using a build of `libnode` from a much older PR revision.


src/NodeApi/JSValue.cs
```diff
@@ -21,38 +20,56 @@ namespace Microsoft.JavaScript.NodeApi;
...
+    /// may cause an attempt to access an invalid memory location.
+    /// </remarks>
     public JSValue(napi_value handle, JSValueScope? scope)
     {
         if (!handle.IsNull && scope is null) throw new ArgumentNullException(nameof(scope));
         _handle = handle;
         _scope = scope;
     }
 
-    public napi_value? Handle
```
vmoroz, line 37: We are removing a check that current Handle is valid. What is an alternative to that check?

src/NodeApi/JSValue.cs
```diff
@@ -167,43 +184,44 @@ public static JSValue SymbolFor(string name)
...
             out napi_value result).ThrowIfFailed(result);
 
     public static unsafe JSValue CreateExternal(object value)
     {
-        GCHandle valueHandle = JSRuntimeContext.Current.AllocGCHandle(value);
+        JSValueScope currentScope = JSValueScope.Current;
+        GCHandle valueHandle = currentScope.RuntimeContext.AllocGCHandle(value);
         return JSValueScope.CurrentRuntime.CreateExternal(
-            Env,
+            CurrentEnv,
```
vmoroz, line 207: Should we use `currentScope` instead?

src/NodeApi/JSValueScope.cs
```diff
@@ -82,25 +82,50 @@ public sealed class JSValueScope : IDisposable
...
-    /// <exception cref="InvalidOperationException">No scope was established for the current
+    /// <exception cref="JSInvalidScopeException">No scope was established for the current
     /// thread.</exception>
     public static JSValueScope Current => s_currentScope ??
-        throw new InvalidOperationException("No current scope.");
+        throw new JSInvalidScopeException(currentScope: null);
+
+    internal int ThreadId { get; }
 
     public bool IsDisposed { get; private set; }
```
vmoroz, line 92: Where do we check that the scope is disposed? It must make any related JSValue invalid.

src/NodeApi/JSValue.cs
```diff
@@ -21,38 +20,56 @@ namespace Microsoft.JavaScript.NodeApi;
 
     internal JSRuntime Runtime => Scope.Runtime;
 
+    /// <summary>
+    /// Creates an empty instance of <see cref="JSValue" />, which implicitly converts to
+    /// <see cref="JSValue.Undefined" /> when used in any scope.
+    /// </summary>
     public JSValue() { }
 
-    public JSValue(napi_value handle) : this(handle, JSValueScope.Current)
```
vmoroz, line 26: This constructor was quite convenient in a number of cases. What is the advantage to remove it?

vmoroz:
I am concerned that PR removes all safety checks from `JSValue`.
After this change the `Handle` is never checked which will cause a lot of security risks of accessing released memory. 
Am I reading this change wrong?

test/JSValueScopeTests.cs
```diff
@@ -0,0 +1,348 @@
...
+        using JSValueScope escapableScope = new(JSValueScopeType.Escapable);
+        Assert.Throws<JSInvalidScopeException>(() =>
+        {
+            using JSValueScope callbackScope = new(JSValueScopeType.Callback);
+        });
+        Assert.Equal(JSValueScopeType.Escapable, JSValueScope.Current.ScopeType);
+    }
+
+    [Fact]
+    public void AccessValueFromClosedScope()
```
vmoroz, line 280: Should we have a test that accesses `Hendle` property from `JSValue`?

Another interesting scenario would be when we have two nested scopes.
One scope creates and object and another one creates a property key.
Then, after the inner scope is closed, we use the key to set object property.
Since the Handle is never checked, the code will access the deleted memory for the property key.

jasongin:
> I am concerned that PR removes all safety checks from `JSValue`. After this change the `Handle` is never checked which will cause a lot of security risks of accessing released memory. Am I reading this change wrong?

The checks are not removed, but moved to the JSValueScope to napi_env conversion. We could have them in both places but I thought that was redundant since you have to get a napi_env to do anything with a handle. I can explain in more detail when I have time later.

test/JSValueScopeTests.cs
jasongin: There is a test for accessing a value from a disposed scope. It is checked.

vmoroz: The test is not covering a case when a function uses two `JSValue`s from different scopes, where one of the is disposed (property key) and the other is not (the object). Since the check for `napi_env` will succeed, the code will use deleted property key `napi_value`.

vmoroz:
> > I am concerned that PR removes all safety checks from `JSValue`. After this change the `Handle` is never checked which will cause a lot of security risks of accessing released memory. Am I reading this change wrong?
> 
> The checks are not removed, but moved to the JSValueScope to napi_env conversion. We could have them in both places but I thought that was redundant since you have to get a napi_env to do anything with a handle. I can explain in more detail when I have time later.

The main goal of `JSValue` is to safely wrap `napi_value`. Replacing it for the `napi_env` check from `JSValueContext` is not equivalent. For the functions that use a single `napi_value`, I agree, there is no difference. But if a function uses two or more `napi_value`, then we effectively check only one of them, while others are used unchecked. It is dangerous if these values are from different scopes.
If we want to get rid of redundant `napi_value`/`napi_env `checks, then we should rather avoid checking for `napi_env` because it will be covered by `napi_value` check, but not the other way around.

src/NodeApi/JSValue.cs
jasongin: I added the checks back to here, so now getting either the `napi_value` or `napi_env` is checked.

src/NodeApi/JSValue.cs
jasongin: Yes, updated.

src/NodeApi/JSValueScope.cs
jasongin: Getting either the `napi_value` or `napi_env` now checks for disposed scope.

test/JSValueScopeTests.cs
jasongin: That is a good point about secondary arguments. I added that test case.

src/NodeApi/JSValue.cs
jasongin: It was only used in one place: the `operator JSValue(napi_value)` in this class. I removed it thinking it might be better to be explicit when using `JSValueScope.Current`, since in this PR I had to fix many places that were incorrectly using `JSValueScope.Current` instead of the scope on the provided `JSValue`. But as long as we still have that operator that is using the current scope implicitly (which we use everywhere) then this constructor is no worse. So I've added it back in the update.

jasongin:
> if a function uses two or more `napi_value`

Yes, that is a good point about additional `napi_value` args.

> If we want to get rid of redundant `napi_value`/`napi_env `checks, then we should rather avoid checking for `napi_env` because it will be covered by `napi_value` check, but not the other way around.

For static functions that take only a `napi_env` and no value handles we still have to validate the thread access to the env via the current scope.

So we'll check in both places, and it will a little redundant sometimes, but it is only a boolean plus an `nint` comparison so it's not a big cost. From benchmarks of scenarios that have heavy interaction with handles, it looks like maybe 1-2% performance difference, though the benchmarks don't quite have that level of precision. Regardless I think is worth it for the added safety.

I suppose we could add a `JSValueScope.UncheckedEnv` handle property that could be used when we know the scope is already being checked via the value, but it's probably not worth the code complexity / risk of bugs.

src/NodeApi/Interop/JSSynchronizationContext.cs
```diff
@@ -7,6 +7,25 @@
...
 namespace Microsoft.JavaScript.NodeApi.Interop;
 
+/// <summary>
+/// Manages the synchronization context for a JavaScript environment, allowing callbacks and
+/// asynchronous continuations to be invoked on the JavaScript thread that runs the environment.
+/// </summary>
+/// <remarks>
+/// All JavaScript values are bound to the thread that runs the JS environment and can only be
+/// accessed from the same thread. Attempts to access a JavaScript value from a different thread
+/// will throw <see cref="JSInvalidScopeException" />.
```
vmoroz, line 17: Can we have a better named exception in such case? E.g. `JSInvalidThreadException`.

src/NodeApi/JSReference.cs
```diff
@@ -165,14 +176,36 @@ T GetValueAndRunAction()
 
     public bool IsDisposed { get; private set; }
 
-    private void ThrowIfDisposed()
+    private void CheckDisposed()
```
vmoroz, line 179: The previous name `ThrowIdDisposed` was much better. It was easy to read and understand. The new name looks "smoother" but difficult to understand without looking at the implementation. It says the completely opposite: it says that it checks that the object is already disposed, but in reality, it checks that it is not disposed.
We should either keep the previous name or rename it to `CheckNotDisposed`.
I would rather keep the existing name unless the .Net SDK has already predefined pattern in that context.

src/NodeApi/JSReference.cs
```diff
@@ -192,14 +225,14 @@ protected virtual void Dispose(bool disposing)
...
             {
                 JSValueScope.CurrentRuntime.DeleteReference(_env, handle).ThrowIfFailed();
             }
             else
             {
-                SynchronizationContext.Post(
-                    () => JSValueScope.CurrentRuntime.DeleteReference(
+                _context.SynchronizationContext.Post(
+                    () => _context.Runtime.DeleteReference(
                         _env, handle).ThrowIfFailed(), allowSync: true);
```
vmoroz, line 236: I wonder why we need the `handle` local variable?
Most probably it was added at some point to avoid capturing `this` in the lambda.
With the current use of other instance fields in the lambda, it may be not necessary anymore.
Though, we may look the other way around and avoid capturing `this`.

src/NodeApi/JSValueScopeClosedException.cs
```diff
@@ -0,0 +1,33 @@
...
+        : base(scope.ScopeType.ToString(), message ?? GetMessage(scope))
+    {
+        Scope = scope;
+    }
+
+    public JSValueScope Scope { get; }
+
+    private static string GetMessage(JSValueScope scope)
+    {
+        return $"The JS value scope of type {scope.ScopeType} was closed.\n" +
```
vmoroz, line 29: Should we also mention the `JSReference`?

vmoroz (approved):
LGTM

src/NodeApi/JSReference.cs
```diff
@@ -165,14 +176,36 @@ T GetValueAndRunAction()
...
+        if ((napi_env)currentScope != _env)
+        {
+            int threadId = Environment.CurrentManagedThreadId;
+            string? threadName = Thread.CurrentThread.Name;
+            string threadDescription = string.IsNullOrEmpty(threadName) ?
+                $"#{threadId}" : $"#{threadId} \"{threadName}\"";
+            string message = "The JS reference cannot be accessed from the current thread.\n" +
+                $"Current thread: {threadDescription}. " +
+                $"Consider using {nameof(JSSynchronizationContext)} to switch to the JS thread.";
+            throw new JSInvalidScopeException(currentScope, message);
```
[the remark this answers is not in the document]
jasongin: This exception type can have other messages:

> There is no active JS value scope. Current thread: {threadDescription}. Consider using JSSynchronizationContext to switch to the JS thread.

> The JS value scope cannot be accessed from the current thread. The scope of type {ScopeType} was created on thread #{threadId} and is being accessed from {threadDescription}.  Consider using JSSynchronizationContext to switch to the JS thread.

So those messages are more about the scope. But in any case the cause is nearly always going to be from using the wrong thread, so I will rename it to `JSInvalidThreadAccessException`.

src/NodeApi/Interop/JSSynchronizationContext.cs
jasongin: Renamed.

src/NodeApi/JSReference.cs
jasongin: I had changed this to be consistent with the new thread access check method, but didn't think carefully about it. I renamed them both to `ThrowIf...`.

src/NodeApi/JSReference.cs
```diff
@@ -76,29 +77,39 @@ public JSReference(napi_ref handle, bool isWeak = false)
...
+            CheckDisposed();
+            CheckThreadAccess();
+            return _env;
+        }
+    }
+
     public void MakeWeak()
     {
-        ThrowIfDisposed();
+        CheckDisposed();
```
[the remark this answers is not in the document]
jasongin: Good point. I removed the redundant checks in this class.

src/NodeApi/JSValueScopeClosedException.cs
jasongin: Good idea. Added.

src/NodeApi/JSReference.cs
jasongin: I checked the history; at the time that variable was added, it was the only thing captured by the lambda. But I don't see any harm in capturing `this`; it might keep the object alive slightly longer, but that shouldn't be an issue. So I removed the extra variable.

jasongin:
In the update I also added consistent public `Handle` and `EnvironmentHandle` properties; the `(napi_value)`, `(napi_env)`, and `(napi_ref)` operators now just call those. That follows framework API design guidelines that any operators should have non-operator equivalent properties or methods. And I added an internal `UncheckedEnvironmentHandle` property that can be used to bypass redundant checks when the scope / env handle will already be checked via value handle access in the same call.

src/NodeApi/JSInvalidScopeException.cs
```diff
@@ -0,0 +1,87 @@
...
+/// </remarks>
+/// <seealso cref="JSSynchronizationContext"/>
+public class JSInvalidScopeException : InvalidOperationException
+{
+    /// <summary>
+    /// Creates a new instance of <see cref="JSInvalidScopeException" /> with a 
+    /// current scope and message.
+    /// </summary>
+    public JSInvalidScopeException(
+        JSValueScope? currentScope,
```
vmoroz, line 28: >JSValueScope? currentScope [](http://example.com/codeflow?start=8&length=26)

Since it is always must be `JSValueScope.Current`, I wonder if we should simplify the constructor and avoid passing it as a parameter, and initialize it from `JSValueScope.Current` in the constructor.

src/NodeApi/JSReference.cs
```diff
@@ -188,19 +237,18 @@ protected virtual void Dispose(bool disposing)
         if (!IsDisposed)
         {
             IsDisposed = true;
-            napi_ref handle = _handle; // To capture in lambda
 
             // The context may be null if the reference was created from a "no-context" scope such
             // as the native host. In that case the reference must be disposed from the JS thread.
-            if (SynchronizationContext == null)
+            if (_context == null)
```
vmoroz, line 243: >_context == null [](http://example.com/codeflow?start=16&length=16)

The `_context` being null is a bit "sketchy". Ideally, `JSReference` must always be associated with a non-null object that wraps up `napi_env`. We must not use or delete the `napi_ref` if that `napi_env` wrapper is disposed.
When `napi_env` is deleted, it also deletes all associated `napi_ref` internally. Any `napi_ref` in `JSReference` must become invalid because they are pointers to deleted data. Calls to `napi_delete_reference` or any other `napi_ref` functions may cause a crash or memory corruption because they would use deleted `napi_env` and/or `napi_ref`.
The `napi_env` disposable wrapper must be for the `JSReference` to be the same as `JSValueScope` for `JSValue` - to protect wrapped handle. It must protect `napi_ref`, while `JSValueScope` protects `napi_value`.
The `napi_env` wrapper must be invalidated in response to finalizer passed to `napi_set_instance_data`.

src/NodeApi/JSValue.cs
```diff
@@ -21,38 +21,87 @@ namespace Microsoft.JavaScript.NodeApi;
...
+                // If the scope is null, this is an empty (uninitialized) instance.
+                // Implicitly convert to the JS `undefined` value.
+                return Undefined._handle;
+            }
+
+            // Ensure the scope is valid and on the current thread (environment).
+            _scope.ThrowIfDisposed();
+            _scope.ThrowIfInvalidThreadAccess();
+
+            // The handle must be non-null when the scope is non-null.
```
vmoroz, line 77: >// The handle must be non-null when the scope is non-null. [](http://example.com/codeflow?start=12&length=58)

Should we add a check in the constructor for it?
Currently we only force scope to be not null if handle is not empty. We never check that handle is not null when handle is not null.
Previously we used to check for null handle: `Handle ?? throw new ...`

src/NodeApi/JSValueScope.cs
```diff
@@ -141,7 +210,7 @@ public sealed class JSValueScope : IDisposable
                 else
                 {
                     throw new InvalidOperationException(
-                        "A Root scope cannot be created within another scope.");
+                        $"A Root scope cannot be created within another scope.");
```
vmoroz, line 213: >$ [](http://example.com/codeflow?start=24&length=1)

nit: why is it needed?

src/NodeApi/JSValueScope.cs
```diff
@@ -174,35 +245,49 @@ public sealed class JSValueScope : IDisposable
...
+                // s_currentScope (which is used to initialize _parentScope above).
+                throw new InvalidOperationException("Parent scope is disposed.");
+            }
+            else if (scopeType == JSValueScopeType.Callback &&
+                _parentScope.ScopeType != JSValueScopeType.Callback &&
+                _parentScope.ScopeType != JSValueScopeType.Module &&
+                _parentScope.ScopeType != JSValueScopeType.Root &&
+                _parentScope.ScopeType != JSValueScopeType.NoContext)
+            {
+                throw new InvalidOperationException(
```
vmoroz, line 270: >InvalidOperationException [](http://example.com/codeflow?start=26&length=25)

The condition above also checks for the NoContext scope type.

vmoroz:
>             throw new InvalidOperationException("Parent scope must not be null.");

Should we use the JSInvalidScopeException in this function?

---
Refers to: src/NodeApi/JSValueScope.cs:390 in c4cc578. [](commit_id = c4cc57839df8677a81095bb955b04a7cf391e04a, deletion_comment = False)

src/NodeApi/JSInvalidScopeException.cs
jasongin: No, because if the current scope is null then `JSValueScope.Current` will throw this exception and that would be infinite recursion.

I'm not sure what happened with the diff: `JSInvalidScopeException.cs` was supposed to be deleted because it was renamed to `JSInvalidThreadAccessException` at your suggestion.

jasongin:
> > ```
> >         throw new InvalidOperationException("Parent scope must not be null.");
> > ```
> 
> Should we use the JSInvalidScopeException in this function?
 
`JSInvalidScopeException` was renamed to `JSThreadAccessException` at your suggestion. So then it didn't make sense to use for the parent scope checks here because they are not about threading. (`JSInvalidScopeException.cs` was mistakenly still included in the dif.)

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -48,15 +50,18 @@ public static napi_value InitializeModule(napi_env env, napi_value exports)
     {
         Trace($"> NativeHost.InitializeModule({env.Handle:X8}, {exports.Handle:X8})");
 
-        JSRuntime runtime = new NodejsRuntime();
-        using JSValueScope scope = new(JSValueScopeType.NoContext, env, runtime);
```
vmoroz, line 52: >JSValueScope [](http://example.com/codeflow?start=14&length=12)

JSValueScoper must be bound to the call stack. Storing it as a field defeats the purpose as it now will say that `napi_value` created in this scope are valid, while in practice they will point to deleted memory.

src/NodeApi/JSInvalidScopeException.cs
vmoroz: I was surprised to see two identical classes with different names. :)

src/NodeApi/JSReference.cs
jasongin: The reason for the `NoContext` scope type is that the `NativeHost` must not set the instance data, because the `ManagedHost` must set instance data to a managed `JSRuntimeContext` instance. User code should never encounter this case because this is a special case for the native host, as commented there.

Currently the native host only creates one reference and keeps it for the process lifetime, so it never actually disposes the reference. So this code never executes when `_context` is null. But I wasn't sure whether the native would ever need to do more things with references in the future. If you prefer, we could change this so that the native host _leaks_ references instead of disposing them, which would be fine if the only reference(s) it uses are process lifetime.

src/NodeApi/DotNetHost/NativeHost.cs
jasongin: I understand and I'll consider a different approach.

This is tricky because the `JSRuntime` instance must be passed to callbacks somehow, and that cannot be through instance data because the `JSRuntime` instance is needed to call the API to get instance data. The alternative is probably to use another thread-local static for the `JSRuntime` instance.

src/NodeApi/JSValueScope.cs
jasongin: Yes, but that scope type is not meant to be used by user code, so I didn't think it was worth mentioning in the exception message. (Root scope may be used in embedding scenarios and Module scope may be used in custom module initializers that don't use the generator, but I don't think NoContext scope is applicable anywhere outside the native host.)

src/NodeApi/JSReference.cs
jasongin: > Ideally, JSReference must always be associated with a non-null object that wraps up `napi_env`.

Currently we wrap the `napi_env` in `JSValueScope`, but that doesn't work for references since they may outlive a scope. I keep thinking maybe we should create a `JSEnvironment` class to wrap the `napi_env`. Then, `JSValueScope`, `JSReference`, and `JSRuntimeContext` could all reference a `JSEnvironment` instance. But I hesitate to create yet another class for maintaining context.

src/NodeApi/JSInvalidScopeException.cs
jasongin: I deleted `JSInvalidScopeException.cs` in an update.

src/NodeApi/JSValue.cs
jasongin: Done.

src/NodeApi/JSValueScope.cs
jasongin: Removed.

src/NodeApi/JSReference.cs
vmoroz: Could we use the `JSRuntimeContext` for that?

src/NodeApi/DotNetHost/NativeHost.cs
vmoroz: Having a thread local for `JSRuntime` does not sound that bad.

jasongin: As discussed, we'll defer this issue for now, and address it in another change soon. If you are adding a `JSEnvironment` class or something like that, it might make sense to use that to resolve this issue. Or if that doesn't work I'll consider the thread-local `JSRuntime` mentioned above. But for now I'll merge this to unblock your work.

src/NodeApi/JSReference.cs
jasongin: I think this is good enough for now, with the added check I added to validate the `_env` matches the current thread-local environment. But I will re-evaluate after the other planned changes to how we track the environment context.

PR #183, opened 2023-11-20
Use new multiline strings and normalize line ends

vmoroz:
The code gen tests were failing on my machine because expected strings had LF line ends while generated strings had environment's CRLF.
In this PR the issue is addressed by calling the `ReplaceLineEndings()` for all expected strings to match current environment line endings.
Also, the expected strings are changed to be the new C# multiline strings.
They allow string text to be aligned with the rest of the code. The indentation is cut by the final `"""` indentation.

### DIFF
--- a/test/TypeDefsGeneratorTests.cs
+++ b/test/TypeDefsGeneratorTests.cs
@@ -46,16 +46,18 @@ public class TypeDefsGeneratorTests
     [Fact]
     public void GenerateSimpleInterface()
     {
-        // NOTE: String literals in these tests use TABS for indendation!
-        Assert.Equal(@"
-/** interface */
-export interface SimpleInterface {
-	/** property */
-	TestProperty: string;
+        // NOTE: String literals in these tests use TABS for indentation!
+        Assert.Equal("""
 
-	/** method */
-	TestMethod(): string;
-}",
+            /** interface */
+            export interface SimpleInterface {
+            	/** property */
+            	TestProperty: string;
+
+            	/** method */
+            	TestMethod(): string;
+            }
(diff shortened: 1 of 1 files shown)


jasongin:
I recommend fixing your git config so the line endings in your source files will be correct on Windows.

```
git config --global core.autocrlf true
```

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

jasongin (approved):

vmoroz:
> I recommend fixing your git config so the line endings in your source files will be correct on Windows.
> 
> ```
> git config --global core.autocrlf true
> ```
> 
> https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

I cannot do it globally. Each OSS project has its own requirements.
We should add repo-specific settings if our code relies on it.

jasongin:
> Each OSS project has its own requirements.

I'm surprised any project would require NOT using that setting, since it is widely recommended for use with git on Windows and I believe the installer for Windows now sets it by default.

Regardless, I'm fine with fixing our tests to be independent of the local newline environment, which is why I approved this PR.

vmoroz:
> > Each OSS project has its own requirements.
> 
> I'm surprised any project would require NOT using that setting, since it is widely recommended for use with git on Windows and I believe the installer for Windows now sets it by default.
> 
> Regardless, I'm fine with fixing our tests to be independent of the local newline environment, which is why I approved this PR.

There are some projects that have heavy dependency on new lines to be LF. Especially if they generate code.  Their developers typically do not use Windows and not aware about the issues the code (mostly test code) might have on Windows machines.

PR #184, opened 2023-11-20
Use released .Net 8 version 8.0.100

vmoroz:
.Net 8 was released last week on 11/14/2023.
This PR updates the `global.json` to use the released .Net 8 version 8.0.100.

### DIFF
--- a/src/NodeApi.Generator/ModuleGenerator.cs
+++ b/src/NodeApi.Generator/ModuleGenerator.cs
@@ -38,7 +38,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
 
     public GeneratorExecutionContext Context { get; protected set; }
 
-    public void Initialize(GeneratorInitializationContext context)
+    public void Initialize(GeneratorInitializationContext _ /*context*/)
     {
         // Note source generators cannot be directly launched in a debugger,
         // because the generator runs at build time, not at application run-time.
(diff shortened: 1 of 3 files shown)


src/NodeApi.Generator/ModuleGenerator.cs
```diff
@@ -38,7 +38,7 @@ public class ModuleGenerator : SourceGenerator, ISourceGenerator
 
     public GeneratorExecutionContext Context { get; protected set; }
 
-    public void Initialize(GeneratorInitializationContext context)
+    public void Initialize(GeneratorInitializationContext _ /*context*/)
```
jasongin, line 41: IDK why this suddenly started generating a warning, and why it doesn't repro locally, only on the build machine. It shouldn't warn since the parameter is defined by the interface. [In my PR](https://github.com/microsoft/node-api-dotnet/pull/182/files#diff-cc7d63bc00415ed9a65524baf55cb0a98d2e74cd2c919203a2a17375a5679907) I suppressed with `#pragma` instead`.

jasongin (approved):
Please also update this section of `publish.yml`:
```
      - task: UseDotNet@2
        displayName: Install .NET SDK using global.json
        inputs:
          packageType: sdk
          useGlobalJson: false # Switch to true and remove explicit version after .NET 8 release
          version: 8.0.x
          includePreviewVersions: true
```

src/NodeApi.Generator/ModuleGenerator.cs
jasongin: Please rebase after #185 was merged; it fixes this.

jasongin (approved):

PR #191, opened 2024-01-19
Update example to SK 1.1, fix TS generator and marshaller bugs

jasongin:
- `examples/semantic-kernel`:
   - Update `Microsoft.SemanticKernel` package reference to `1.1.0`
   - Update JS code for SK API changes.
 - TS generator fixes:
   - Improve the logic for merging nuget package references with system assembly references.
   - Omit explicit interface implementations for value types (as was already done for classes).
   - Omit members with pointer parameter types.
   - Convert `<see cref="..."/>` to JSDoc `{@link ...}`
 - Marshalling fixes:
   - Fix an exception caused by certain kinds of circular references among exported types.
   - Enable marshalling primitive JS value types to .NET `object` type. Necessary to support some SK APIs such as [KernelArguments(IDictionary<string, object?> source)](https://github.com/microsoft/semantic-kernel/blob/b2bbd410c07739767b5aad48961a8ced5b4cf4a4/dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs#L47C49-L47C49) where the expectation is a JS string value in a `Map` should get marshalled to a NET string value in the dictionary.
 - Tracing improvements:
   - Automatically turn on tracing when a debugger is attached.
   - Enable callback tracing for getters, setters, and methods. Previously the callback tracing only worked for regular functions.

### DIFF
--- a/examples/semantic-kernel/example.js
+++ b/examples/semantic-kernel/example.js
@@ -4,11 +4,12 @@
 // @ts-check
 
 import dotnet from 'node-api-dotnet';
+import './bin/System.Text.Encodings.Web.js';
+import './bin/Microsoft.Extensions.DependencyInjection.js';
 import './bin/Microsoft.Extensions.Logging.Abstractions.js';
 import './bin/Microsoft.SemanticKernel.Abstractions.js';
 import './bin/Microsoft.SemanticKernel.Core.js';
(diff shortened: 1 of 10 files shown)


vmoroz (approved):
LGTM

PR #192, opened 2024-01-20
Defer dynamic export of type members until first access

jasongin:
Previously, when dynamically invoking .NET APIs from JS (as in the `semantic-kernel` example) loading of types within an assembly/namespace was deferred, but whenever a type was loaded, marshalling code was generated and compiled for all members of the type, and all types referenced by any of the members, cascading on to the referenced types' members and so on. This meant for a nontrivial application thousands of types and members could be processed at startup time, the vast majority of which were never actually used by the app.

This change defers the marshalling code generation/compilation for members (constructors, properties, and methods) of loaded types, until the first time each individual member is accessed. And then, there is no need to load referenced types until that time either.

This change has two major benefits:
 1. It significantly improves startup time for the dynamic invocation scenario. (The SK example runs over 2X faster, not counting the network call.)
 2. It prevents any marshalling code-gen bugs with APIs deep in the type dependency tree (unused by the application) from blocking the application startup. While there has been a lot of work to fix crashes while processing all different kinds of .NET APIs, whether they are fully supported for marshalling or not, there are probably a few edge cases still lurking.

Summary of changes:
 - In `TypeExporter` when exporting properties, initially define the property with getter/setter callbacks that load the actual property getter/setter and then redefine the property before calling it.
 - Enhance `JSCallbackOverload` to support deferred loading of overloaded constructor/method callbacks, using a `Lazy<T>` to invoke the loader once (which generates and compiles the marshalling code).
 - In `TypeExporter`, use `JSCallbackOverload` to defer loading of constructors and methods.
    - For non-overload constructors and methods this may have a slight perf impact since there is now an additional level of indirection. With more work it could be possible to redefine the method properties (but not the constructors). But it might not be worth the extra code.

Related diagnostic improvements:
 - Update tracing related to exporting type members.
 - Rename existing `TRACE` and `DEBUG` environment variables to use a consistent `NODE_API_` prefix, and update `README-dev.md`.
 - Make the new delayed export behavior conditional on a `NODE_API_DELAYLOAD` variable, enabled by default. (I'm not sure it's needed, but it's possible it could cause some unforeseen problems.)

### DIFF
--- a/README-DEV.md
+++ b/README-DEV.md
@@ -53,12 +53,20 @@ with no CLR.
 ## Debugging
 With a debug build, the following environment variables trigger just-in-time debugging of the
 respective components:
- - `DEBUG_NODE_API_GENERATOR` - Debug the C# source-generator when it runs during the build.
- - `DEBUG_NODE_API_RUNTIME` - Debug the .NET runtime host when it is loaded by JavaScript. (Does
+ - `NODE_API_DEBUG_GENERATOR=1` - Debug the C# source-generator or TS type-definitions generator
+ when they runs during the build.
+ - `NODE_API_DEBUG_RUNTIME=1` - Debug the .NET runtime host when it is loaded by JavaScript. (Does
  not apply to AOT-compiled modules.)
+Setting either of these variables to `1` causes the program to print a message to the console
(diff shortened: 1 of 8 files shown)


vmoroz (approved):
LGTM

PR #196, opened 2024-02-02
Implement strongly typed cast from/to JSValue

vmoroz:
This PR implements checked and unchecked casting from/to `JSValue` to other types that wrap `JSValue`.
- `JSValue` has new generic methods: `Is<T>`, `As<T>`, `AsUnchecked<T>`, and `CastTo<T>`.
- These methods are implemented by using the new interface `IJSValue` that has static methods `CanCreateFrom` and `CreateUnchecked`. In case if we targeted .Net version is below version 7.0, it uses `IJSValueShim` helper class that binds to static methods using reflection.
- The types that wrap up the `JSValue` implement these static methods.
- Explicit casting operators are changed to use the new `JSValue.As<T>` method.
- New unit tests are added to test the new casting operations.

Please note that I could not find a good way to differentiate `Proxy` from `Object` and currently any `JSValue` that is an object can be casted to `JSProxy`.

### DIFF
--- a/src/NodeApi/IJSValue.cs
+++ b/src/NodeApi/IJSValue.cs
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#if !NET7_0_OR_GREATER
+using System;
+using System.Reflection;
+#endif
+
+namespace Microsoft.JavaScript.NodeApi;
+
+#if NET7_0_OR_GREATER
+// A static interface that helps with the conversion of JSValue to a specific type.
+public interface IJSValue<TSelf> where TSelf : struct, IJSValue<TSelf>
+{
+    public static abstract bool CanBeConvertedFrom(JSValue value);
+
+    public static abstract TSelf CreateUnchecked(JSValue value);
+}
+#else
+// A static class that helps with the conversion of JSValue to a specific type.
+public static class IJSValueShim<T> where T : struct
+{
+    private static readonly Func<JSValue, bool> s_canBeConvertedFrom =
+        (Func<JSValue, bool>)Delegate.CreateDelegate(
+            typeof(Func<JSValue, bool>),
+            typeof(T).GetMethod(
+                nameof(JSObject.CanBeConvertedFrom),
+                BindingFlags.Static | BindingFlags.Public)!);
+
+    private static readonly Func<JSValue, T>s_createUnchecked =
+        (Func<JSValue, T>)Delegate.CreateDelegate(
+            typeof(Func<JSValue, T>),
+            typeof(T).GetMethod(
+                nameof(JSObject.CreateUnchecked),
+                BindingFlags.Static | BindingFlags.Public)!);
+
+    public static bool CanBeConvertedFrom(JSValue value) => s_canBeConvertedFrom(value);
+
+    public static T CreateUnchecked(JSValue value) => s_createUnchecked(value);
+}
+#endif
--- a/src/NodeApi/IJSValue.cs
+++ b/src/NodeApi/IJSValue.cs
@@ -1,40 +1,45 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#if !NET7_0_OR_GREATER
 using System;
+
+#if !NET7_0_OR_GREATER
 using System.Reflection;
 #endif
 
 namespace Microsoft.JavaScript.NodeApi;
 
-#if NET7_0_OR_GREATER
 // A static interface that helps with the conversion of JSValue to a specific type.
-public interface IJSValue<TSelf> where TSelf : struct, IJSValue<TSelf>
+public interface IJSValue<TSelf> : IEquatable<JSValue> where TSelf : struct, IJSValue<TSelf>
 {
-    public static abstract bool CanBeConvertedFrom(JSValue value);
+    public JSValue AsJSValue();
+
+#if NET7_0_OR_GREATER
+    public static abstract bool CanCreateFrom(JSValue value);
 
     public static abstract TSelf CreateUnchecked(JSValue value);
+#endif
 }
-#else
+
+#if !NET7_0_OR_GREATER
 // A static class that helps with the conversion of JSValue to a specific type.
-public static class IJSValueShim<T> where T : struct
+internal static class IJSValueShim<T> where T : struct
 {
-    private static readonly Func<JSValue, bool> s_canBeConvertedFrom =
+    private static readonly Func<JSValue, bool> s_canBeCreatedFrom =
         (Func<JSValue, bool>)Delegate.CreateDelegate(
             typeof(Func<JSValue, bool>),
             typeof(T).GetMethod(
-                nameof(JSObject.CanBeConvertedFrom),
+                "CanCreateFrom",
                 BindingFlags.Static | BindingFlags.Public)!);
 
     private static readonly Func<JSValue, T>s_createUnchecked =
         (Func<JSValue, T>)Delegate.CreateDelegate(
             typeof(Func<JSValue, T>),
             typeof(T).GetMethod(
-                nameof(JSObject.CreateUnchecked),
-                BindingFlags.Static | BindingFlags.Public)!);
+                "CreateUnchecked",
+                BindingFlags.Static | BindingFlags.NonPublic)!);
 
-    public static bool CanBeConvertedFrom(JSValue value) => s_canBeConvertedFrom(value);
+    public static bool CanCreateFrom(JSValue value) => s_canBeCreatedFrom(value);
 
     public static T CreateUnchecked(JSValue value) => s_createUnchecked(value);
 }
--- a/src/NodeApi/IJSValue.cs
+++ b/src/NodeApi/IJSValue.cs
@@ -18,17 +18,47 @@ namespace Microsoft.JavaScript.NodeApi;
 public interface IJSValue<TSelf> : IEquatable<JSValue> where TSelf : struct, IJSValue<TSelf>
 {
     /// <summary>
-    /// Converts the derived struct `TSelf` to a <see cref="JSValue"/> with
-    /// the same `napi_value` handle.
+    /// Checks if the T struct can be created from this instance`.
     /// </summary>
+    /// <typeparam name="T">A struct that implements IJSValue interface.</typeparam>
     /// <returns>
-    /// <see cref="JSValue"/> with the same `napi_value` handle as the derived struct.
+    /// `true` if the T struct can be created from this instance. Otherwise it returns `false`.
     /// </returns>
-    JSValue AsJSValue();
+    bool Is<T>() where T : struct, IJSValue<T>;
+
+    /// <summary>
+    /// Tries to create a T struct from this instance.
+    /// It returns `null` if the T struct cannot be created.
+    /// </summary>
+    /// <typeparam name="T">A struct that implements IJSValue interface.</typeparam>
+    /// <returns>
+    /// Nullable value that contains T struct if it was successfully created
+    /// or `null` if it was failed.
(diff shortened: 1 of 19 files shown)


src/NodeApi/JSObject.cs
```diff
@@ -9,11 +9,16 @@
...
+#endif
 {
     private readonly JSValue _value;
 
-    public static explicit operator JSObject(JSValue value) => new(value);
-    public static implicit operator JSValue(JSObject obj) => obj._value;
+    public static implicit operator JSValue(JSObject value) => value.AsJSValue();
+    public static explicit operator JSObject?(JSValue value) => value.As<JSMap>();
+    public static explicit operator JSObject(JSValue value)
+        => value.As<JSObject>() ?? throw new InvalidCastException("JSValue is not an Object.");
```
jasongin, line 21: Should there be a generic method that does the equivalent of this checked cast? How about `JSValue.CastTo<TValue>()`?

(Note we have non-generic "CoerceTo" methods already.)

src/NodeApi/JSObject.cs
```diff
@@ -24,6 +29,16 @@ public JSObject() : this(JSValue.CreateObject())
     {
     }
 
+    #region IJSValue<JSObject> implementation
+
+    public static bool CanBeConvertedFrom(JSValue value) => value.IsObject();
```
jasongin, line 34: Can static interface methods be explicitly implemented? I ask because these methods are an implementation detail that might be better to hide; the naming is a bit awkward and we want to encourage use of the `Is<>()` / `As<>()` methods instead.

src/NodeApi/IJSValue.cs
```diff
@@ -0,0 +1,41 @@
...
+#if !NET7_0_OR_GREATER
+using System;
+using System.Reflection;
+#endif
+
+namespace Microsoft.JavaScript.NodeApi;
+
+#if NET7_0_OR_GREATER
+// A static interface that helps with the conversion of JSValue to a specific type.
+public interface IJSValue<TSelf> where TSelf : struct, IJSValue<TSelf>
```
jasongin, line 13: Should the `AsJSValue()` method be included in this interface? All the value types have that method anyway.

src/NodeApi/IJSValue.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public static abstract bool CanBeConvertedFrom(JSValue value);
```
jasongin, line 15: Should this method be named `CanCreateFrom()`? It doesn't really do any "conversion" of the value.

src/NodeApi/IJSValue.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public static abstract TSelf CreateUnchecked(JSValue value);
+}
+#else
+// A static class that helps with the conversion of JSValue to a specific type.
+public static class IJSValueShim<T> where T : struct
```
jasongin, line 21: This class can be `internal`. Or you could make the delegates private members of `JSValue`.

src/NodeApi/IJSValue.cs
jasongin, line 13: Consider making this interface extend `IEquatable<JSValue>`, which is already implemented by all the value types.

src/NodeApi/JSObject.cs
```diff
@@ -9,11 +9,16 @@
 namespace Microsoft.JavaScript.NodeApi;
 
 public readonly partial struct JSObject : IDictionary<JSValue, JSValue>, IEquatable<JSValue>
+#if NET7_0_OR_GREATER
+    , IJSValue<JSObject>
+#endif
```
jasongin, line 14: Maybe the `IJSValue` interface could still be defined for < .NET 7, just without the static interface methods. Then this condition wouldn't be needed.

src/NodeApi/JSObject.cs
vmoroz: Added `JSValue.CastTo<T>`

src/NodeApi/IJSValue.cs
vmoroz: The `AsJSValue()` is added to the IJSValue interface.

src/NodeApi/IJSValue.cs
vmoroz: Renamed to `CanCreateFrom`

src/NodeApi/IJSValue.cs
vmoroz: Changed to be `internal` They cannot be part of `JSValue`. The delegates must be per `T`.

src/NodeApi/IJSValue.cs
vmoroz: Done.

src/NodeApi/JSObject.cs
vmoroz: It is changed per the suggestion.

src/NodeApi/JSObject.cs
vmoroz: Yep, they can be explicitly implemented and I made the `CanCreateFrom` static method private.

src/NodeApi/IJSValue.cs
```diff
@@ -1,41 +1,80 @@
...
+    /// </summary>
+    /// <returns>
+    /// <see cref="JSValue"/> with the same `napi_value` handle as the derived struct.
+    /// </returns>
+    JSValue AsJSValue();
 
-    public static abstract TSelf CreateUnchecked(JSValue value);
+#if NET7_0_OR_GREATER
+    /// <summary>
+    /// Checks id the derived struct `TSelf` can be created from a <see cref="JSValue"/>.
```
jasongin, line 31: typo

src/NodeApi/JSValue.cs
```diff
@@ -391,29 +391,58 @@ public unsafe BigInteger ToBigInteger()
...
 
-    public bool Is<TValue>() where TValue : struct
+    /// <summary>
+    /// Checks if the T struct can be created from this `JSValue`.
+    /// </summary>
+    /// <typeparam name="T">A struct that implements IJSValue interface.</typeparam>
+    /// <returns>
+    /// `true` if the T struct can be created from this `JSValue`. Otherwise it returns `false`.
+    /// </returns>
+    public bool Is<T>() where T : struct, IJSValue<T>
```
jasongin, line 401: Should these methods (`Is<>`, `As<>`, `AsUnchecked<>`, `CastTo<>`) be default interface implementations of `IJSValue<>` instead of methods of `JSValue`? Then they could convert _between_ specific types, not only from `JSValue` to a specific type. For example it is valid to convert a `JSArray` or `JSPromise` to `JSObject`.

src/NodeApi/JSValue.cs
jasongin, line 1: Should `JSValue` also implement `IJSValue<JSValue>`?

src/NodeApi/JSValue.cs
```diff
@@ -391,29 +391,58 @@ public unsafe BigInteger ToBigInteger()
...
+    /// It throws `InvalidCastException` in case of failure.
+    /// </summary>
+    /// <typeparam name="T">A struct that implements IJSValue interface.</typeparam>
+    /// <returns>T struct created based on this `JSValue`.</returns>
+    /// <exception cref="InvalidCastException">
+    /// Thrown when the T struct cannot be crated based on this `JSValue`.
+    /// </exception>
+    public T CastTo<T>() where T : struct, IJSValue<T>
+        => As<T>()
+        ?? throw new InvalidCastException("JSValue cannot be casted to target type.");
```
jasongin, line 444: Include the target type `typeof(T).Name` in the message.

test/TestCases/napi-dotnet/jsvalue_cast.js
jasongin, line 1: It would have been be simpler to write all these tests in C# using the libnode embedding model. But that won't work until we have libnode packages published. (Without that the embedding test cases don't run in CI.)

src/NodeApi/IJSValue.cs
```diff
@@ -1,41 +1,80 @@
...
+/// Implements IJSValue interface static functions for the previous .Net versions.
+/// </summary>
+/// <typeparam name="T"></typeparam>
+internal static class IJSValueShim<T> where T : struct, IJSValue<T>
 {
-    private static readonly Func<JSValue, bool> s_canBeConvertedFrom =
+    /// <summary>
+    /// A static field to keep a reference to the CanCreateFrom private method.
+    /// </summary>
+    private static readonly Func<JSValue, bool> s_canBeCreatedFrom =
```
jasongin, line 53: nit: `s_canCreateFrom`

src/NodeApi/IJSValue.cs
```diff
@@ -1,41 +1,80 @@
...
         (Func<JSValue, bool>)Delegate.CreateDelegate(
             typeof(Func<JSValue, bool>),
             typeof(T).GetMethod(
-                nameof(JSObject.CanBeConvertedFrom),
-                BindingFlags.Static | BindingFlags.Public)!);
+                "CanCreateFrom",
```
jasongin, line 57: `nameof(CanCreateFrom)` since there is now a method of that name in the current class.

src/NodeApi/IJSValue.cs
```diff
@@ -1,41 +1,80 @@
...
+    /// <summary>
+    /// A static field to keep a reference to the CreateUnchecked private method.
+    /// </summary>
     private static readonly Func<JSValue, T>s_createUnchecked =
         (Func<JSValue, T>)Delegate.CreateDelegate(
             typeof(Func<JSValue, T>),
             typeof(T).GetMethod(
-                nameof(JSObject.CreateUnchecked),
-                BindingFlags.Static | BindingFlags.Public)!);
+                "CreateUnchecked",
```
jasongin, line 67: `nameof(CreateUnchecked)`

src/NodeApi/JSValue.cs
jasongin: This assumes `JSValue` also implements `IJSValue<JSValue>` as suggested above.

jasongin (approved):

src/NodeApi/IJSValue.cs
vmoroz: Fixed

src/NodeApi/JSValue.cs
vmoroz: Unfortunately the default interface implementations are not accessible from the struct or classes that implement it. I have added these methods to all structs and now we can convert between all types that implement `IJSValue`.
As a result, the `AsJSValue()` method became redundant and I had removed it.

src/NodeApi/JSValue.cs
vmoroz: Good idea! Done.

src/NodeApi/JSValue.cs
vmoroz: Done

test/TestCases/napi-dotnet/jsvalue_cast.js
vmoroz: Creating the libNode is the next topic on my TODO list.

src/NodeApi/IJSValue.cs
vmoroz: Fixed.

src/NodeApi/IJSValue.cs
vmoroz: Fixed.

src/NodeApi/IJSValue.cs
vmoroz: Fixed.

src/NodeApi/JSValue.cs
jasongin: The message is not updated.

vmoroz: Fixed in the latest commit.

Issue #206, opened 2024-02-07
byte[] from .NET becomes string in Node


james-hu:
I am calling .NET from Node. 
There's a .NET method returning byte[]. In the generated type definition, the return type is Number[].
However, the actul data type of the object returned (`typeof x`) is 'string'.
Length of the string (number of characters in the string) is is shorter than the length of the byte array (number of elements in the array). I I've tried hacking into it, but I cound't find a way to convert the return value correctly to Buffer or byte array for consuming.

Is this a bug?

jasongin:
Probably a bug. I think we don't have good tests for this scenario right now. I'll have to investigate.

Issue #64, opened 2023-03-29
Custom marshalling


jasongin:
API proposal at https://github.com/jasongin/napi-dotnet/pull/25

PR #220, opened 2024-03-09
Fix byte-array JSValue conversions

jasongin:
Fixes: #206 

A .NET API that has a `byte[]` (in a property type, parameter type, or return type) should get projected as JS `number[]` type, _not_ a JS `string` type. This change removes the conversion operators between `byte[]` and string-type `JSValue`. Since code-generation for type conversions makes use of those operators, removing those conversions achieves the desired effect of keeping the array form of the type.

In some scenarios it might actually be desirable to marshal .NET `byte[]` as JS `string`. So, it could be helpful then if there was a way for the application to control that behavior (related to #64). But I believe `number[]` is the most correct default.

There were a few places in test code that were relying on the implicit conversion from `byte[]` to `JSValue`; I replaced those with explicit conversions.


vmoroz:
Previously, the `byte[]` conversion to JS string was done specifically to support the C# UTF-8 strings.
My concern is that this PR stops us from using UTF-8 strings.
Could we have an alternative way to convert `byte[]` to a number array while keeping the existing UTF-8 string support?

vmoroz (approved):

PR #223, opened 2024-03-11
Fix spelling for the IsCanceled property

vmoroz:
This is a tiny change to fix spelling of the `IsCanceled` property.


jasongin (approved):
Two L's in "cancelled" are technically not wrong, though I think one L is the preferred American spelling, and matches `Task.IsCanceled`.

PR #247, opened 2024-03-24
Fix marshalling of classes that override Equals

jasongin:
Updates/supersedes PR #156 

Class objects are marshalled by reference and therefore must use reference equality in the object map. Previously, a class overrides `Equals()` to implement value quality semantics would get confused by the marshaller (object map) with other instances that were value-equal.

Structs are marshalled by value, not tracked in the object map, and are not guaranteed to maintain reference equality after marshalling.


DmitriySft:
Thank you for this work. The PR looks good.

vmoroz (approved):

PR #261, opened 2024-04-06
Fix for types derived from un-namespaced types.

DmitriySft:
```Type.Namespace``` can be null. There are cases where such cases are not correctly handled.

The easiest example is:

``` CSharp
interface INoNamespaceInterface { }
namespace MyNamespace
{
    public class MyType: INoNamespaceInterface {} // Gives NRE
}
```

This PR fixes this problem.

But codebase-wise, there are places where ```Type.Namespace == null``` is not handled well. Suggest to fix all of them together with this feature request https://github.com/microsoft/node-api-dotnet/issues/262


DmitriySft:

src/NodeApi.DotNetHost/TypeExporter.cs
```diff
@@ -212,7 +212,14 @@ private void RegisterDerivedType(TypeProxy derivedType, Type? baseOrInterfaceTyp
         {
             string baseOrInterfaceTypeName = TypeProxy.GetTypeProxyName(baseOrInterfaceType);
 
+            if (baseOrInterfaceType.Namespace == null)
+            {
+                Trace($"Skipping un-namespaced base or interface: {baseOrInterfaceType.Name}");
+                return;
+            }
+
             NamespaceProxy? ns = GetNamespaceProxy(baseOrInterfaceType.Namespace!);
```
jasongin, line 221: Did you consider updating `GetNamespaceProxy()` to handle a null namespace value instead? It could simply use an empty string instead when the value is null. I think that might be better, considering it will need to support (not skip) un-namespaced types soon anyway for #262.

DmitriySft: Ready

jasongin:
Thanks for the PR and bug report.

It looks like the CI is just complaining about some pedantic issues in the test code. For info about that check see https://github.com/microsoft/node-api-dotnet/blob/main/README-DEV.md#checkfix-formatting

Since it's test code, it would be OK to disable some of those warnings with `#pragma warning disable`

jasongin (approved):

Issue #216, opened 2024-03-06
JSThreadSafeFunction APIs should not use ThrowIfFailed()


jasongin:
The `JSThreadSafeFunction` class has public methods such as `BlockingCall()` and `NonBlockingCall()` that are documented as callable from any thread, not only the JS thread. Those methods use the internal `ThrowIfFailed()` method to check the status code. But that method may construct an instance of `JSError()`, which uses several APIs that must be called from the JS thread.

The result is a `JSThreadSafeFunction` failure can result in a `JSInvalidThreadAccessException` instead of an error about the actual failure. I found this because the error can be _occasionally_ reproduced with the `napi-dotnet/streams` test case. That test may fail with the following stack trace:
```
Unhandled Exception: Microsoft.JavaScript.NodeApi.JSInvalidThreadAccessException: There is no active JS value scope.
Current thread: #2. Consider using the synchronization context to switch to the JS thread.
   at Microsoft.JavaScript.NodeApi.JSValueScope.get_Current()
   at Microsoft.JavaScript.NodeApi.JSErrorInfo.GetLastErrorInfo()
   at Microsoft.JavaScript.NodeApi.JSError..ctor(String message, JSErrorType errorType, String code)
   at Microsoft.JavaScript.NodeApi.NodeApiStatusExtensions.ThrowIfFailed(napi_status status, String memberName, String sourceFilePath, Int32 sourceLineNumber)
   at Microsoft.JavaScript.NodeApi.Interop.JSThreadSafeFunction.NonBlockingCall(napi_status status)
   at Microsoft.JavaScript.NodeApi.Interop.JSThreadSafeFunction.NonBlockingCall(Action callback)
   at Microsoft.JavaScript.NodeApi.Interop.JSTsfnSynchronizationContext.Post(SendOrPostCallback callback, Object state)
   at Microsoft.JavaScript.NodeApi.Interop.JSSynchronizationContext.Post(Action action, Boolean allowSync)
   at Microsoft.JavaScript.NodeApi.JSReference.Dispose(Boolean disposing)
   at Microsoft.JavaScript.NodeApi.JSReference.Finalize()
```
When I debugged I found that the failed status code was `napi_closing`, which makes sense because it is being called from a finalizer.

Issue #285, opened 2024-04-23
Visual Studio build ModuleGenerator warning 


Waleed-KH:
When building a .NET Node Module project in Visual Studio, it fail to generate the marshalling code

```
CSC : warning CS8032: An instance of analyzer Microsoft.JavaScript.NodeApi.Generator.ModuleGenerator cannot be created from C:\Users\...\.nuget\packages\microsoft.javascript.nodeapi.generator\0.7.8\analyzers\dotnet\cs\Microsoft.JavaScript.NodeApi.Generator.dll : Could not load file or assembly 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
```

Visual Studio Version 17.9.6

jasongin:
Yes, currently the generator targets .NET 6.0 or later, so it doesn't work with how VS runs MSBuild and the C# compiler on .NET Framework 4.x.

I am working on switching the generator to target .NET Standard 2.0, as is recommended for source-generators, but there are some issues I need to work through.

Waleed-KH:
fixed #290

PR #290, opened 2024-04-26
Make the source-generator target .NET Standard 2.0

jasongin:
Fixes: #216, #285

When building inside Visual Studio, VS runs MSBuild and the C# compiler on .NET Framework 4.x -- unlike with `dotnet build` where they run on the .NET 6+ runtime matching the current .NET SDK, even when building a project targeting .NET Framework 4.x. That means analyzer assemblies (which the NodeApi source-generator is) must support the same .NET runtime that the compiler is using in either case. The general guidance is to build analyzers to target `netstandard2.0`.

So we have to add a `netstandard2.0` target framework to all the library projects and make most of the code compatible. Fortunately since it was already compatible with both .NET Framework 4.x and .NET 6+, it wasn't too difficult to additionally support .NET Standard 2.0.

Then it gets a little more complicated since we also launch the same generator assembly during the build as an executable tool for generating typedefs, and _that_ should match the target framework of the project being built, which may be different from the current MSBuild runtime version. Some new logic in `Microsoft.NodeApi.Generator.targets` handles that.

After all the changes I found that some tests were hitting bug #216 more frequently on .NET Framework 4.x, so I did a simple fix for that also (though we may consider better error-handling there later).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
-    <TargetFrameworks Condition=" '$(TargetFrameworks)' == '' and ! $([MSBuild]::IsOsPlatform('Windows')) ">net8.0;net6.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(TargetFrameworks)' == '' and $([MSBuild]::IsOsPlatform('Windows')) ">net8.0;net6.0;net472</TargetFrameworks>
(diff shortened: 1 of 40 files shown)


vmoroz:
> #endif // !NETFRAMEWORK

nit: please rename. The same seems to be in some other files.

---
Refers to: src/NodeApi/Interop/JSCollectionProxies.cs:376 in 75015c2. [](commit_id = 75015c29fb2621e6b2bc8633c9caadf8110063e2, deletion_comment = False)

vmoroz (approved):
:shipit:

Issue #292, opened 2024-05-07
NAOT: nullable and exceptions repeated calls


daniacedue:
Greetings,

I have found a bug which seems to originate from the code generator.
To replicate the bug we need a class which has a nullable property to another class and another method which throws an exception (which we catch in Node JS). Looping through these calls will corrupt something in the generated code somehow.

Here the source code of an example:
```
using Microsoft.JavaScript.NodeApi;
namespace ExceptionTest
{
    [JSExport]
    public class ExceptionGreeter
    {
        public ExceptionGreeter()
        {
            SubGreeter = new SubGreeter();
        }
        public SubGreeter? SubGreeter { get; set; }
    }
    [JSExport]
    public class SubGreeter
    {
        public void ThrowException(string msg)
        {
            throw new Exception(msg);
        }
    }
}
```
and on the Node JS side (using typescript):
```
import {ExceptionGreeter} from "../ExceptionTest/publish-debug/ExceptionTest.js";
const exgr = new ExceptionGreeter();
for (let i = 0; i<100; i++){
    try{
        exgr.subGreeter?.throwException('exception ' + i);
    }catch(err){
        console.log(err);
    }
}
```
In this example the loop will not reach 100, on my machine it will stop at exactly 54 loops each time ending with an unhandled exception:
```
Unhandled Exception: Microsoft.JavaScript.NodeApi.JSException: Invalid argument: Error in DefineProperties at D:\a\1\s\src\NodeApi\JSValue.cs:535
   at file:///C:/REPOS3/microsof-nodeapi.tests/ExceptionTest/nodeapp/main.js:5:14
   at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
   at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
   at async loadESM (node:internal/process/esm_loader:28:7)
   at async handleMainPromise (node:internal/modules/run_main:120:12)
   at Microsoft.JavaScript.NodeApi.NodeApiStatusExtensions.ThrowIfFailed(JSRuntime.napi_status, String, String, Int32) + 0x13c
   at Microsoft.JavaScript.NodeApi.JSValue.<>c__DisplayClass108_0.<DefineProperties>b__0(String _, ReadOnlySpan`1 descriptorsPtr) + 0x95
[...]

PR #294, opened 2024-05-11
Fix crash due to released weakref in object map

jasongin:
Fixes: #292

When repeatedly passing the same .NET object to JS, and then releasing the JS objects (but not the .NET object), the internal .NET-to-JS object mapping code could encounter an exception due to incorrect handling of the weak JS reference.

 - Add a `JSReference.HasValue` property that checks if a reference is both not disposed and has an available value.
 - Add doc-comments on `JSReference` methods about when `ObjectDisposedException` may be thrown.
 - Fix object mapping code working with the JS weak references, and reduce code duplication.

### DIFF
--- a/src/NodeApi/JSReference.cs
+++ b/src/NodeApi/JSReference.cs
@@ -54,6 +54,7 @@ public class JSReference : IDisposable
     /// <summary>
     /// Gets the value handle, or throws an exception if access from the current thread is invalid.
     /// </summary>
+    /// <exception cref="ObjectDisposedException">The reference is disposed.</exception>
     /// <exception cref="JSInvalidThreadAccessException">Access to the reference is not valid on
     /// the current thread.</exception>
     public napi_ref Handle
@@ -110,6 +111,10 @@ public class JSReference : IDisposable
         }
     }
 
+    /// <summary>
+    /// Changes a strong reference into a weak reference, if it is not already weak.
+    /// </summary>
+    /// <exception cref="ObjectDisposedException">The reference is disposed.</exception>
     public void MakeWeak()
     {
         if (!IsWeak)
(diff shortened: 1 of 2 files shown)


src/NodeApi/JSReference.cs
```diff
@@ -131,18 +142,26 @@ public void MakeStrong()
...
         JSValueScope.CurrentRuntime.GetReferenceValue(Env, _handle, out napi_value result)
             .ThrowIfFailed();
         return result;
     }
 
+    /// <summary>
+    /// Gets a value indicating whether the referenced JS value is still available (not released).
+    /// Returns false (does not throw) if the reference is disposed.
+    /// </summary>
+    public bool HasValue => !IsDisposed && (!IsWeak || GetValue().HasValue);
```
vmoroz, line 157: I have seen many times how the `HasValue`-like property is misused in code and often causes issues. E.g. in the code developer may check `HasValue`, get a positive result, and act upon it, but the value is then destroyed in a different thread. A much better approach is to request a strong value and act on it depending if it is null or not.
Should we just add the `IsDisposed` check into the `GetValue()` method and use it instead?

vmoroz (approved):

src/NodeApi/JSReference.cs
jasongin: > check `HasValue`, get a positive result, and act upon it, but the value is then destroyed in a different thread

That isn't a concern with this API because references can only be accessed from a single thread.

I have also given feedback about mis-use of similar "check" APIs (such as `ContainsKey()` or `File.Exists()`) in many code reviews. But that doesn't invalidate the utility of such APIs in some legitimate use cases, as in the places where I'm using `HasValue` in this PR.

Without multithreading, there is still a minor issue of inefficiency if the API is mis-used. But that's no different from many other equally valid APIs.

> Should we just add the `IsDisposed` check into the `GetValue()` method and use it instead?

I think it is correct for `GetValue()` to throw `ObjectDisposedException` when the reference is disposed. That can help catch some bugs.

I don't have a problem with making the `HasValue` property `internal` if you prefer. But I thought it could be generally helpful in some valid use cases.

PR #295, opened 2024-05-12
Remove undefined behavior from JSTypedArray.GetJSValueForMemory

DaZombieKiller:
The `GetJSValueForMemory` method currently uses unsafe code to access the internals of the `Memory<T>` type, resulting in undefined behavior.

Contrary to the comments inside, there has always been a public API to access the `MemoryManager<T>` backing a `Memory<T>` instance: `MemoryMarshal.TryGetMemoryManager<T, TManager>`. This PR replaces the implementation with one using that API to remove the undefined behavior.

With this new implementation, it is no longer necessary to strip the high bit from the index (it is an implementation detail that the `MemoryMarshal` APIs do not expose). Technically it wasn't necessary to begin with as that bit is only set for `Memory<T>` instances backed by pinned arrays, which are already ruled out by checking for a `MemoryManager` instance.

### DIFF
--- a/src/NodeApi/JSTypedArray.cs
+++ b/src/NodeApi/JSTypedArray.cs
@@ -106,26 +106,13 @@ public readonly struct JSTypedArray<T> : IEquatable<JSValue> where T : struct
     /// returns that JS value.
     /// </summary>
     /// <returns>The JS value, or null if the memory is external to JS.</returns>
-    private static unsafe JSValue? GetJSValueForMemory(ReadOnlyMemory<T> data)
+    private static JSValue? GetJSValueForMemory(ReadOnlyMemory<T> data)
     {
-        // This assumes the owner object of a Memory struct is stored as a reference in the
-        // first (private) field of the struct. If the Memory internal structure ever changes
-        // (in a future major version of the .NET Runtime), this unsafe code could crash.
-        // Unfortunately there's no public API to get the Memory owner object.
-        void* memoryPointer = Unsafe.AsPointer(ref data);
-        object? memoryOwner = Unsafe.Read<object?>(memoryPointer);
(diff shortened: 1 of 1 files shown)


DaZombieKiller:

jasongin:
Thanks for the PR! This is a great clean-up, as the previous code there (that I wrote) was certainly dubious. I guess I just did not discover the `TryGetMemoryManager()` API.

src/NodeApi/JSTypedArray.cs
```diff
@@ -106,25 +106,10 @@ public unsafe JSTypedArray(ReadOnlyMemory<T> data)
...
-            // The Memory was created from a JS TypedArray.
-
-            // Strip the high bit of the index - it has a special meaning.
-            void* memoryIndexPointer = (byte*)memoryPointer + Unsafe.SizeOf<object?>();
-            int index = Unsafe.Read<int>(memoryIndexPointer) & ~int.MinValue;
-
-            void* memoryLengthPointer = (byte*)memoryIndexPointer + Unsafe.SizeOf<int>();
-            int length = Unsafe.Read<int>(memoryLengthPointer);
-
             JSValue value = manager.JSValue;
```
jasongin, line 113: The CI build reports a warning on this line:

> Dereference of a possibly null reference.

Can you fix the warning? I think adding a `!` will be sufficient. I don't know why that is necessary since the `out manager` parameter does have `[NotNullWhen(true)]`.

jasongin (approved):

src/NodeApi/JSTypedArray.cs
DaZombieKiller: That is very strange, I wonder if the CI is compiling against reference assemblies that lack the nullability attributes or something? I'll add the suppression now.

PR #325, opened 2024-07-15
Disable CodeQL for OSX ARM64 and remove debug logging

vmoroz:
In this PR we disable CodeQL for OSX ARM64 until the issue of using x64 bits on OSX ARM64 platform is fixed.
This PR also removes the debugging code introduced in the previous PR #324.

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -45,25 +45,15 @@ jobs:
 
       variables:
         DotNetMoniker: net8.0
-        # Temporary variables to collect CodeQL info to debug OSX ARM64 build issue.
-        # They must be removed as soon as they not needed to save storage.
+        # Temporary disable CodeQL for OSX ARM64.
+        # It must be restored after CodeQL fixes the issue where they install x64 bits on OSX ARM64.
         ${{ if eq( matrixEntry.TargetRuntime, 'osx-arm64' ) }}:
-          Codeql.PublishDatabaseLog: true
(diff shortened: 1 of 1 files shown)


.ado/publish.yml
```diff
@@ -45,25 +45,15 @@ jobs:
...
       variables:
         DotNetMoniker: net8.0
-        # Temporary variables to collect CodeQL info to debug OSX ARM64 build issue.
-        # They must be removed as soon as they not needed to save storage.
+        # Temporary disable CodeQL for OSX ARM64.
+        # It must be restored after CodeQL fixes the issue where they install x64 bits on OSX ARM64.
         ${{ if eq( matrixEntry.TargetRuntime, 'osx-arm64' ) }}:
-          Codeql.PublishDatabaseLog: true
-          Codeql.Cadence: 0
+          Codeql.Enabled: false
```
jasongin, line 51: I didn't think this would fix this issue because from what I read the injection still happens, it just returns immediately. Since the problem is with loading the injection, I think it will still fail.

jasongin (approved):

.ado/publish.yml
vmoroz: Hmm. Let me still try it. This fix was suggested by CodeQL team today.

vmoroz: The CI build is passed. So, the fix works. :)

Issue #197, opened 2024-02-03
JS reference API improvements


jasongin:
Currently we have a [`JSReference`](https://github.com/microsoft/node-api-dotnet/blob/main/src/NodeApi/JSReference.cs) class (that is a wrapper for a [`napi_ref`](https://nodejs.org/api/n-api.html#napi_ref) handle), which can be either a strong or weak reference.

I have noticed this pattern is very common when writing code that works with strong references:
 
```C#
someReference.GetValue()!.Value
```
 
The node.js doc says [napi_get_reference_value](https://nodejs.org/api/n-api.html#napi_get_reference_value) may return NULL if the JS reference is not "still valid". I think that implies it can only return NULL if the reference is weak. (We should check what happens if the environment was disposed; I'm assuming it returns an error rather than a NULL value handle.)

When working with strong references, the non-null assumption should be OK. But for cleaner code there maybe should be a separate API for getting the value from a strong reference. It could also check the assumption that the reference is actually strong.

One possible approach could be to add a separate `JSWeakReference` type, where `JSWeakReference.GetValue()` returns a nullable value while `JSReference.GetValue()` returns a non-nullable value (or else throws). There are some considerations with that:
  - The `MakeWeak()` / `MakeStrong()` methods would return instances of the opposite type, and might have to dispose the original instance to avoid confusion. Or we could omit the capability to promote/demote an existing reference, and instead allocate a new reference to return from those methods.
  - JavaScript already defines a [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) type that enables JS code to work with weak references. The distinction between them is subtle; a `WeakRef` is itself a JS object, and then you could have a  (weak or strong) `JSReference` to that value. But I can't imagine a scenario when one would need to work with both of those at the same time.

jasongin:
I expect that the vast majority of `JSReference` uses will be strong references. Using weak references is a more advanced scenario, so I think it's generally OK to expect code to be aware of the possibility of an unavailable value in that case.

Proposal:

 - Keep the single `JSReference` type.
 - Add a `TryGetValue(out JSValue value)` method:
    - Returns `true` and the value if it succeeds.
    - Returns `false` and `default(JSValue)` if the reference is weak and the JS value is unavailable.
    - Throws `ObjectDisposedException` if the reference is disposed.
 - Change the `GetValue()` method return type to a non-nullable `JSValue`, instead of the current nullable return type.
[...]

PR #326, opened 2024-07-15
JS Reference API improvements

jasongin:
Fixes: #197 

 - Add a `JSReference.TryGetValue(out JSValue value)` method:
    - Returns `true` and the value if it succeeds.
    - Returns `false` and `default(JSValue)` if the reference is weak and the JS value is unavailable.
    - Throws `ObjectDisposedException` if the reference is disposed.
 - Change the `JSReference.GetValue()` method return type to a non-nullable `JSValue`, instead of the previous nullable return type.
    - Returns the value if it succeeds.
    - Throws `NullReferenceException` if the reference is weak and the JS value is unavailable.
    - Throws `ObjectDisposedException` if the reference is disposed.
 - Replace many instances of `GetValue()!.Value` with `GetValue()` in the codebase.
 - Update documentation.

I considered also adding generic variants of the methods above, to return more specific JS value types, but that will have to wait for #196 which adds the generic value type conversions.

### DIFF
--- a/src/NodeApi/Interop/JSRuntimeContext.cs
+++ b/src/NodeApi/Interop/JSRuntimeContext.cs
@@ -311,7 +311,7 @@ public sealed class JSRuntimeContext : IDisposable
 
         if (_objectMap.TryGetValue(obj, out JSReference? existingWrapperWeakRef))
         {
-            if (!existingWrapperWeakRef.IsDisposed && existingWrapperWeakRef.GetValue().HasValue)
+            if (!existingWrapperWeakRef.IsDisposed && existingWrapperWeakRef.TryGetValue(out _))
             {
                 // If the .NET object is already mapped to a non-released JS object, then
                 // another one should not be created.
@@ -626,6 +626,9 @@ public sealed class JSRuntimeContext : IDisposable
 
         JSReference reference = _importMap.GetOrAdd((module, property), (_) =>
         {
+            // TODO: Consider what to do if the imported property is undefined
+            // or not a referenceable type.
+
             if (module == null || module == "global")
             {
                 // Importing a built-in object from `global`.
@@ -656,9 +659,8 @@ public sealed class JSRuntimeContext : IDisposable
                     return new JSReference(propertyValue);
                 }
             }
-
         });
-        return reference.GetValue() ?? JSValue.Undefined;
+        return reference.GetValue();
     }
 
     /// <summary>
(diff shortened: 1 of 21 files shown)


src/NodeApi/Interop/JSRuntimeContext.cs
```diff
@@ -656,9 +659,8 @@ private JSValue GetOrCreateObjectWrapper<T>(T obj, Func<JSValue> createWrapper)
                     return new JSReference(propertyValue);
                 }
             }
-
         });
-        return reference.GetValue() ?? JSValue.Undefined;
+        return reference.GetValue();
```
vmoroz, line 663: It seems that the logic is changed here: we used to return `Undefined` if the reference has no value, and now it throws.

src/NodeApi/Interop/NodeStream.cs
```diff
@@ -18,7 +18,7 @@ public partial class NodeStream : Stream
 
     public static explicit operator NodeStream(JSValue value) => new(value);
     public static implicit operator JSValue(NodeStream stream)
-        => stream._valueReference.GetValue() ?? default;
```
vmoroz, line 21: It is another place where we start to throw.

src/NodeApi/JSReference.cs
```diff
@@ -28,8 +28,15 @@ public class JSReference : IDisposable
...
 
-    public bool IsWeak { get; private set; }
-
+    /// <summary>
+    /// Creates a new instance of a <see cref="JSReference"/> that holds a strong or weak
+    /// reference to a JS value.
+    /// </summary>
+    /// <param name="value">The JavaScript value to reference.</param>
+    /// <param name="isWeak">True if the reference will be "weak", meaning the reference does not
+    /// prevent the value from being released and garbage-collected. The default is false,
```
vmoroz, line 37: Should we say that GC in this context is JS GC, not .Net GC?

vmoroz (approved):

src/NodeApi/Interop/JSRuntimeContext.cs
jasongin: It is always a strong reference so it could never have returned `Undefined` before.

src/NodeApi/Interop/NodeStream.cs
jasongin: It is always a strong reference so it could never have returned null before.

src/NodeApi/JSReference.cs
jasongin: Updated.

Issue #242, opened 2024-03-22
Marshalling: Collection classes like List<T>, Dictionary<T>, ReadOnlyCollection<T>


jasongin:
.NET collection _interfaces_ like `IList<T>`, `IDictionary<T>`, `IEnumerable<T>` etc. are already supported for marshalling to/from JS collections. This issue is about adding support for marshalling collections where the property or parameter type is defined as a _class_ like `List<T>`, `Dictionary<T>`, or `ReadOnlyCollection<T>` instead of an interface.

Use of collection classes in public APIs is generally considered a bad practice:
 - [Why we don't recommend using List<T> in public APIs](https://learn.microsoft.com/en-us/archive/blogs/kcwalina/why-we-dont-recommend-using-listt-in-public-apis)
 - [CA1002 - Do not expose generic lists](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1002)

Still, a lot of code does this anyway, especially if the APIs are ultimately internal to an application and not part of a polished public SDK. So, it makes sense to have some support for marshalling collection classes.

Unfortunately there are some limitations that make efficient marshalling of collection classes difficult. While .NET collection interfaces are marshalled by reference (proxy) in both directions, classes cannot use the same approach because there is no way to fully override the implementation of `List<T>` or `Dictionary<T>` in a way that replaces the internal representation.

There was an attempt to implement this feature in #128, which stalled due to this challenge.

When revisiting this, we'll need to carefully consider the semantics for marshalling all variations of collections and arrays in .NET and JS and decide on an approach that is as consistent and efficient as possible, while dealing with some unavoidable compromises.

rijulg:
Just a note on this issue considering I worked on the initial PR. I ultimately ended up extending the external dependency I wanted to support and improved typing on that which allowed me to circumvent having the collection support in the node-api-dotnet library.

Issue #243, opened 2024-03-22
Marshalling: Non-generic collections


jasongin:
.NET generic collection interfaces like `IList<T>`, `IDictionary<T>`, `IEnumerable<T>` are already supported for marshalling to/from equivalent JS collections.

This issue is about adding support for non-generic collections `IList`, `IDictionary`, `IEnumerable` and possibly some others.

I don't think non-generic collections are used much anymore, but maybe we should consider supporting them for completeness.

See also #242 about generic collection _classes_. We should also consider non-generic collection classes as part of this issue.

PR #344, opened 2024-08-01
Collections marshalling improvements

jasongin:
Fixes: #242

 - Marshal the collection classes in the `System.Collections.Generic` namespace, such as `List<>`, `Queue<>`, `Stack<>`, `HashSet<>`, and `Dictionary<>`.
   - Since these classes are all `sealed` they cannot be overridden to proxy collection data access to JS. Therefore the collections are marshalled by value, similar to .NET arrays.
 - Marshal some collection classes in the `System.Collections.ObjectModel` namespace, such as `Collection<>` and `ReadOnlyDictionary<>`.
   - These classes can be marshalled by reference, similar to the generic collection interfaces, because they are already designed to wrap/proxy access to other collections.
   - `ObservableCollection<>` is not supported for now. It could be possible to support observability in JS, but that will depend on events support (#59).
  - Add a `JSPropertyDescriptor.ToObject()` conversion and use it with `JSProxy.GetOwnPropertyDescriptor` to allow for more convenient proxying of that method.
  - Fix a bug in iteration of JS collection proxies - it required additional proxy traps.
  - Update the doc page about collections.

Non-generic collections are still not supported. (#243)


vmoroz (approved):

src/NodeApi/JSPropertyDescriptor.cs
```diff
@@ -68,6 +79,22 @@ namespace Microsoft.JavaScript.NodeApi;
         Data = data;
     }
 
+    /// <summary>
+    /// Creates a property descriptor with a value.
+    /// </summary>
+    public static JSPropertyDescriptor Property(
```
vmoroz, line 85: I find the new name to be less descriptive than the previous `ForValue` name. I would suggest to use the same term as in the JS docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty
The use terms "data descriptor" vs "accessor descriptor".
E,g, we can do the following renames:
- `Property` -> `DataProperty`
- `Accessor` -> `AccessorProperty`
- `Function` -> `Method`

jasongin: Yeah I debated a few different options there.

I changed `ForValue` because it seemed awkward when the other methods (`Accessor` and `Function`) don't begin with `For`. The reason for the `For` prefix is that there is already a property named `Value`, so there cannot be a method with the same name.

`DataProperty` and `AccessorProperty` are good ideas. But `Method` won't work because there is already a property of the class named `Method`.

Alternatively we could just make these all overloads of the `JSPropertyDescriptor` constructor. It would be more concise, but less obvious what kind of descriptor is being created.

vmoroz: We can keep the `Function` then. IMHO, the static methods are more descriptive than constructors.

jasongin: Done.

PR #373, opened 2024-09-03
fix: use intersection for `DateTime`, `DateTimeOffset` type definitions

Waleed-KH:
According to the docs, `DateTime` & `DateTimeOffset` get mapped to JS `Date` value by **adding** either a `kind` or `offset` property to a regular `Date` object. For more accurate type definitions, it should use intersection (`&`) instead of union (`|`)


Waleed-KH:

jasongin (approved):
Thanks!

PR #380, opened 2024-09-18
Node.js worker thread API

jasongin:
- Add `NodeWorker` class to wrap the [Node.js worker thread API](https://nodejs.org/api/worker_threads.html)
   - At the moment there is a limitation in that it doesn't support JS/.NET interop in the worker thread. (The JS worker can still communicate with the .NET host via messages.) Direct interop in the worker will require some of the changes in the `libnode` embedding API that @vmoroz is working on now, specifically to allow the embedding host to provide modules.
 - Add `NodeProcess` class to wrap a small subset of the Node.js `process` API, which is useful for workers (mostly for C# interop in workers which is not done yet).
 - Add `JSObjectDictionary` adapter that supports adapting a JS object to an `IDictionary` with value conversions.
   - It's similar to the `Map` dictionary adapter except JS objects have slightly different semantics than maps.
   - This is used with the `NodeProcess.Env` to project the env JS object as `IDictionary<string, string>`.
 - Fix a bug in `NodeStream.ReadAsync()`, and improve validation in the `NodeStream` constructor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -156,6 +156,9 @@ csharp_space_between_square_brackets = false
 dotnet_diagnostic.CA1510.severity = none
 dotnet_diagnostic.CA1513.severity = none
 
+# Stream APIs with Memory parameters are not available in .NET Framework
+dotnet_diagnostic.CA1835.severity = none
+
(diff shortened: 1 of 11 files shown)


vmoroz (approved):
:shipit:

PR #422, opened 2025-01-22
Add .NET 9 support, remove .NET 6

jasongin:
- Update `global.json` to .NET 9 instead of .NET 8.
 - Replace the .NET 6 target everywhere with .NET 9
   - Keep supporting .NET 8 and .NET Framework 4.7.2
   - .NET 8 is still the default for some things since 8 is LTS and 9 is not.
 - Fix some minor compatibility issues with .NET 9 SDK.
 - Use Node.js v22 instead of v20 in build pipelines.
   - Keep testing with Node.js v18 also.
 - Update package version from 0.8 to 0.9

At this point all the tests are passing, but I want to make sure the example projects are in good shape too before checking this in.


vmoroz (approved):

PrestigeDevop:
any plan to support node v23 for typescript native support ?

PR #425, opened 2025-02-11
LibNode Discovery

wasabii:
Resubmitting a PR I did directly against @vmoroz's local fork earlier. As we talked about over there, right now, this PR has the LibNode packages pointing to my local fork, where the runtimes/{rid}/native directory is populated correctly. The PR thus is not ready to be accepted until the upstream LibNode packages are fixed up. But getting the discovery code visible I thought would be beneficial.

Allow null to be passed as libpath. The result is that a series of procedures for locating the libnode library is done.

a) On Core+ platforms, NativeLibrary is used directly with a relative path off of the assembly. This causes a search across the applications .deps.json. This is the ideal scenario as long as the LibNode packages deliver the library into runtimes/{rid}/native.
b) On Framework platforms a search procedure is used. First, libnode is attempted to be loaded by relative name directly. This handles the situation where the library is directly available in some way. Such as relative to the executable. Second runtimes/{rid}/native is examined with a known RID based on current platform.

A few local NativeLibrary methods were improved.

+ Load was updated to support non-Windows platforms. This handles the scenario of Framework on ~Windows. Such as Mono.
+ TryLoad was introduced to probe paths during discovery.
+ GetExport/TryGetExport was updated to support non-Windows platforms. This handles the scenario of Framework on ~Windows. Such as Mono.
+ dlerror() was introduced to handle the situation of null exports. Which should not reaaaaly happen in libnode. But is technically correct.
+ The pinvokes to libdl are technically out of date. dl was merged into libc on Linux. And was already present in libc on OS X. But these are left.
+ Expects to use Microsoft.JavaScript.LibNode version that deposit native libraries properly.
Change tests to not need a hard coded libnode path.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,7 @@
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" /><!-- 4.1.0 is compatible with .NET Standard -->
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.202-alethic.4" />
+    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.203" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.5.0" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <PackageVersion Include="Nerdbank.GitVersioning" Version="3.6.133" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,7 @@
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" /><!-- 4.1.0 is compatible with .NET Standard -->
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.202" />
+    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.202-alethic.4" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.5.0" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <PackageVersion Include="Nerdbank.GitVersioning" Version="3.6.133" />
--- a/bench/Benchmarks.cs
+++ b/bench/Benchmarks.cs
@@ -83,7 +83,6 @@ public abstract class Benchmarks
     protected void Setup()
     {
         NodeEmbeddingPlatform platform = new(
-            null,
             new NodeEmbeddingPlatformSettings { Args = s_settings });
 
         // This setup avoids using NodejsEmbeddingThreadRuntime so benchmarks can run on
--- a/bench/Benchmarks.cs
+++ b/bench/Benchmarks.cs
@@ -44,12 +44,6 @@ public abstract class Benchmarks
             .WithOptions(ConfigOptions.JoinSummary));
     }
 
-    private static string LibnodePath { get; } = Path.Combine(
-        GetRepoRootDirectory(),
-        "bin",
-        GetCurrentPlatformRuntimeIdentifier(),
-        "libnode" + GetSharedLibraryExtension());
-
     private NodeEmbeddingRuntime? _runtime;
     private NodeEmbeddingNodeApiScope? _nodeApiScope;
     private JSValue _jsString;
@@ -89,7 +83,7 @@ public abstract class Benchmarks
     protected void Setup()
     {
         NodeEmbeddingPlatform platform = new(
-            LibnodePath,
+            null,
             new NodeEmbeddingPlatformSettings { Args = s_settings });
 
         // This setup avoids using NodejsEmbeddingThreadRuntime so benchmarks can run on
--- a/src/NodeApi/Runtime/NativeLibrary.cs
+++ b/src/NodeApi/Runtime/NativeLibrary.cs
@@ -6,7 +6,6 @@
 using System;
 using System.ComponentModel;
 using System.Runtime.InteropServices;
-using System.Runtime.InteropServices.ComTypes;
 
 namespace Microsoft.JavaScript.NodeApi.Runtime;
 
--- a/src/NodeApi/Runtime/NativeLibrary.cs
+++ b/src/NodeApi/Runtime/NativeLibrary.cs
@@ -141,7 +141,7 @@ public static class NativeLibrary
         {
             dlerror();
             procAddress = dlsym(handle, name);
-            return dlerror() != 0;
+            return dlerror() == 0;
         }
 #else
         return SysNativeLibrary.TryGetExport(handle, name, out procAddress);
--- a/src/NodeApi/Runtime/NativeLibrary.cs
+++ b/src/NodeApi/Runtime/NativeLibrary.cs
@@ -44,9 +44,47 @@ public static class NativeLibrary
     public static nint Load(string libraryName)
     {
 #if NETFRAMEWORK || NETSTANDARD
-        return LoadLibrary(libraryName);
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return LoadLibrary(libraryName);
+        }
+        else
+        {
+            var h = dlopen(libraryName, RTLD_LAZY);
+            if (h == 0)
+                throw new DllNotFoundException();
+
+            return h;
+        }
 #else
         return SysNativeLibrary.Load(libraryName);
+#endif
+    }
+
+    /// <summary>
+    /// Provides a simple API for loading a native library and returns a value that indicates whether the operation succeeded.
+    /// </summary>
+    /// <param name="libraryName">The name of the native library to be loaded.</param>
+    /// <param name="handle">When the method returns, the OS handle of the loaded native library.</param>
+    /// <returns><c>true</c> if the native library was loaded successfully; otherwise, <c>false</c>.</returns>
+    public static bool TryLoad(string libraryName, out nint handle)
+    {
+        if (libraryName is null)
+            throw new ArgumentNullException(nameof(libraryName));
+
+#if NETFRAMEWORK || NETSTANDARD
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            handle = LoadLibrary(libraryName);
+            return handle != 0;
+        }
+        else
+        {
+            handle = dlopen(libraryName, RTLD_LAZY);
+            return handle != 0;
+        }
+#else
+        return SysNativeLibrary.TryLoad(libraryName);
 #endif
     }
 
@@ -58,8 +96,28 @@ public static class NativeLibrary
     /// <returns>The address of the symbol.</returns>
     public static nint GetExport(nint handle, string name)
     {
+        if (handle == 0)
+            throw new ArgumentNullException(nameof(handle));
+        if (name is null)
+            throw new ArgumentNullException(nameof(name));
+
 #if NETFRAMEWORK || NETSTANDARD
-        return GetProcAddress(handle, name);
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return GetProcAddress(handle, name);
+        }
+        else
+        {
+            // clear any existing errors
+            dlerror();
+
+            var address = dlsym(handle, name);
+            var error = dlerror();
+            if (error != 0)
+                throw new EntryPointNotFoundException(Marshal.PtrToStringAuto(error));
+
+            return address;
+        }
 #else
         return SysNativeLibrary.GetExport(handle, name);
 #endif
@@ -68,8 +126,20 @@ public static class NativeLibrary
     public static bool TryGetExport(nint handle, string name, out nint procAddress)
     {
 #if NETFRAMEWORK || NETSTANDARD
-        procAddress = GetProcAddress(handle, name);
-        return procAddress != default;
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            procAddress = GetProcAddress(handle, name);
+            return procAddress != default;
+        }
+        else
+        {
+            // clear any existing errors
+            dlerror();
+
+            procAddress = dlsym(handle, name);
+            var error = dlerror();
+            return error != 0;
+        }
 #else
         return SysNativeLibrary.TryGetExport(handle, name, out procAddress);
 #endif
@@ -86,7 +156,27 @@ public static class NativeLibrary
     [DllImport("kernel32")]
     private static extern nint GetProcAddress(nint hModule, string procName);
 
-    private static nint dlopen(nint fileName, int flags)
+    private static nint dlerror()
+    {
+        // Some Linux distros / versions have libdl version 2 only.
+        // Mac OS only has the unversioned library.
+        try
+        {
+            return dlerror2();
+        }
+        catch (DllNotFoundException)
+        {
+            return dlerror1();
+        }
+    }
+
+    [DllImport("libdl", EntryPoint = "dlerror")]
+    private static extern nint dlerror1();
+
+    [DllImport("libdl.so.2", EntryPoint = "dlerror")]
+    private static extern nint dlerror2();
+
+    private static nint dlopen(string fileName, int flags)
     {
         // Some Linux distros / versions have libdl version 2 only.
         // Mac OS only has the unversioned library.
@@ -101,10 +191,30 @@ public static class NativeLibrary
     }
 
     [DllImport("libdl", EntryPoint = "dlopen")]
-    private static extern nint dlopen1(nint fileName, int flags);
+    private static extern nint dlopen1(string fileName, int flags);
 
     [DllImport("libdl.so.2", EntryPoint = "dlopen")]
-    private static extern nint dlopen2(nint fileName, int flags);
+    private static extern nint dlopen2(string fileName, int flags);
+
+    private static nint dlsym(nint handle, string name)
+    {
+        // Some Linux distros / versions have libdl version 2 only.
+        // Mac OS only has the unversioned library.
+        try
+        {
+            return dlsym2(handle, name);
+        }
+        catch (DllNotFoundException)
+        {
+            return dlsym1(handle, name);
+        }
+    }
+
+    [DllImport("libdl", EntryPoint = "dlsym")]
+    private static extern nint dlsym1(nint fileName, string flags);
+
+    [DllImport("libdl.so.2", EntryPoint = "dlsym")]
+    private static extern nint dlsym2(nint fileName, string flags);
 
     private const int RTLD_LAZY = 1;
 
--- a/src/NodeApi/Runtime/NativeLibrary.cs
+++ b/src/NodeApi/Runtime/NativeLibrary.cs
@@ -133,7 +133,7 @@ public static class NativeLibrary
         {
             nint procAddress = GetProcAddress(handle, name);
             if (procAddress == 0 && throwOnError)
-                throw new DllNotFoundException(new Win32Exception(Marshal.GetLastWin32Error()).Message);
+                throw new EntryPointNotFoundException(new Win32Exception(Marshal.GetLastWin32Error()).Message);
 
             return procAddress;
         }
--- a/src/NodeApi/Runtime/NativeLibrary.cs
+++ b/src/NodeApi/Runtime/NativeLibrary.cs
@@ -6,6 +6,7 @@
 using System;
 using System.ComponentModel;
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
 
 namespace Microsoft.JavaScript.NodeApi.Runtime;
 
@@ -146,23 +147,30 @@ public static class NativeLibrary
     [DllImport("kernel32", SetLastError = true)]
     private static extern nint GetProcAddress(nint hModule, string procName);
 
+    private delegate nint DlErrorDelegate();
+    private static DlErrorDelegate? s_dlerror;
+
     private static nint dlerror()
     {
+        // cache dlerror function
+        if (s_dlerror is not null)
+            return s_dlerror();
+
         // some operating systems have dlerror in libc, some in libdl, some in libdl.so.2
         // attempt in that order
         try
         {
             return dlerror0();
         }
-        catch (DllNotFoundException)
+        catch (EntryPointNotFoundException)
         {
             try
             {
-                return dlerror1();
+                return (s_dlerror = dlerror1)();
             }
             catch (DllNotFoundException)
             {
-                return dlerror2();
+                return (s_dlerror = dlerror2)();
             }
         }
     }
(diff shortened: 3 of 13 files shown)


Directory.Packages.props
```diff
@@ -7,7 +7,7 @@
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" /><!-- 4.1.0 is compatible with .NET Standard -->
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.202" />
+    <PackageVersion Include="Microsoft.JavaScript.LibNode" Version="20.1800.202-alethic.4" />
```
vmoroz, line 10: No, we cannot use your package.

wasabii: I know that. Presumably you would fix your own to be .NET compatible.

vmoroz: Ah, I see. I did not read the description before looking through the code. Sorry.

bench/Benchmarks.cs
```diff
@@ -89,7 +83,7 @@ public static void Method() { }
     protected void Setup()
     {
         NodeEmbeddingPlatform platform = new(
-            LibnodePath,
+            null,
```
vmoroz, line 86: Since it becomes optional, it is better to remove this parameter and instead make it a property for the `NodeEmbeddingPlatformSettings`

Directory.Packages.props
wasabii: Yes. This should be accessible https://github.com/orgs/alethic/packages?repo_name=Microsoft.JavaScript.LibNode

And this was the GHA run. The packages would also be in the nuget artifact: https://github.com/alethic/Microsoft.JavaScript.LibNode/actions/runs/13220509513

The build procedure that I put together is in the same repo

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -44,12 +44,50 @@ public static nint GetMainProgramHandle()
...
     {
 #if NETFRAMEWORK || NETSTANDARD
-        return LoadLibrary(libraryName);
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return LoadLibrary(libraryName);
+        }
+        else
+        {
+            var h = dlopen(libraryName, RTLD_LAZY);
```
vmoroz, line 53: I bet our formatter will complain about the `var`. Also, it is preferrable to use full names like `handle` or `libHandle`.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -44,12 +44,50 @@ public static nint GetMainProgramHandle()
...
 #endif
     }
 
+    /// <summary>
+    /// Provides a simple API for loading a native library and returns a value that indicates whether the operation succeeded.
+    /// </summary>
+    /// <param name="libraryName">The name of the native library to be loaded.</param>
+    /// <param name="handle">When the method returns, the OS handle of the loaded native library.</param>
+    /// <returns><c>true</c> if the native library was loaded successfully; otherwise, <c>false</c>.</returns>
+    public static bool TryLoad(string libraryName, out nint handle)
```
vmoroz, line 70: Can we implement the `Load` by using the `TryLoad`? It is to reduce the code duplication.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -58,8 +96,28 @@ public static nint Load(string libraryName)
     /// <returns>The address of the symbol.</returns>
     public static nint GetExport(nint handle, string name)
     {
+        if (handle == 0)
+            throw new ArgumentNullException(nameof(handle));
+        if (name is null)
```
vmoroz, line 101: `name` cannot be `null` here. Should we rather use the `IsNullOrEmpty`?

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -58,8 +96,28 @@ public static nint Load(string libraryName)
...
+            throw new ArgumentNullException(nameof(name));
+
 #if NETFRAMEWORK || NETSTANDARD
-        return GetProcAddress(handle, name);
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return GetProcAddress(handle, name);
```
vmoroz, line 107: I wonder if we should also add the error handing for Windows platform here too.
If you cannot do it, then please add a TODO here.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -68,8 +126,20 @@ public static nint GetExport(nint handle, string name)
     public static bool TryGetExport(nint handle, string name, out nint procAddress)
```
vmoroz, line 126: Can we implement the `GetExport` in terms of the `TryGetExport`? Or we loose ability the for the error reporting?

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -86,7 +156,27 @@ public static bool TryGetExport(nint handle, string name, out nint procAddress)
...
-    private static nint dlopen(nint fileName, int flags)
+    private static nint dlerror()
+    {
+        // Some Linux distros / versions have libdl version 2 only.
+        // Mac OS only has the unversioned library.
+        try
+        {
+            return dlerror2();
+        }
+        catch (DllNotFoundException)
```
vmoroz, line 167: It sounds quite inefficient to throw on each `GetExport` function call,
Can we do the binding only once and cache it somewhere?

Ideally, we must stop using the `DllImport` completely since it is reflection-based.
I wonder if the new `LibraryImport` based on Roslyn code gen would work for the old .Net Framework versions.

wasabii: Yeah. I didn't like this. I was just copying what ya'll did though.

I believe the proper answer at this point in history is to import it from libc, however. libdl exists for compatibility purposes only in my understanding.

wasabii: I looked this up a few years ago, and I think that's the case in every supported Linux distribution supported by MS to at least .NET 3.1. But my memory may be foggy on this. But I went through it a year ago.....

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: I really just copied this bit from Microsoft's NativeLibrary implementation upstream.

But the null check exists in upstream's NativeLibrary because nullable variables in C# are a compiler feature, and are erased at runtime. So, yes, technically, it can still be null.

Looks like you call it through NodeJSRuntime.Import, which also isn't null checked.... I can't trace this far back enough to determine if a null can get through a public contract of any kind.

So, up to you.

Directory.Packages.props
vmoroz: Perfect! I can access it form the artifacts. Thank you! 
I should be able to produce a new Nuget package tomorrow if everything goes well.
The Linux build for ARM64 takes about 4 hours. So, with the PR builds, code reviews, and then CI builds I cannot do it faster.

wasabii: Works for me. Kinda crazy though. Takes 30 minutes on GHA. Wonder if I left something out.

src/NodeApi/Runtime/NativeLibrary.cs
vmoroz: Lets avoid the empty string names. I see more issues with them rather than any benefits in this code.
Also, it should avoid the formatter complains.

Directory.Packages.props
vmoroz: You are not compiling for Linux ARM64 as far I can see.
In my case it is rather complex script:
```
                        cd nodejs
                        sudo apt install binutils-aarch64-linux-gnu gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
                        export CC=aarch64-linux-gnu-gcc
                        export CXX=aarch64-linux-gnu-g++
                        export CC_host="gcc"
                        export CXX_host="g++"
                        ./configure --shared --dest-cpu=arm64 --cross-compiling --dest-os=linux --with-arm-float-abi=hard --with-arm-fpu=neon
                        make -j4
```

Hmm... I guess I must start with removing the `j4` :)

vmoroz: It took me some time walking through the Node.js build system to figure it out.
I am using the `ubuntu-latest` as the OS image.

vmoroz: BTW, a couple of questions about your Nuget package contents:
- Why do we need the `_._` file in the `lib\netstandard2.0` folder?
- Does the `buildTransitive\net461` apply to all .Net Framework version or only to 4.61?

wasabii: My linux-arm64 build is running on GHA's ubuntu-24.04-arm image. Which is linux-arm64, and should be building for linux-arm64. And the .so is arm64:

file libnode.so

libnode.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=8f19436cc4063264c32ab8affbafab85b2a19941, with debug_info, not stripped

wasabii: Ahh. I'm building on a native ARM64 host. So I don't need xcompile options.

wasabii: ADO never got around to adding hosted build agents for ARM. Been like a years since GHA had em.

wasabii: I think both of us have made a mistake here by not using proper sysroots. I have versioned imports for GLIBC_2.38 and a few other things which would need to be fixed for proper compatibility.

wasabii: Okay. More in depth examination of your build output:

```
$ objdump  -T libnode.so
0000000000000000      DF *UND*  0000000000000000 (GLIBC_2.34) pthread_rwlock_trywrlock
0000000000000000      DF *UND*  0000000000000000 (GLIBC_2.2.5) memset
0000000000000000      DF *UND*  0000000000000000 (GLIBC_2.34) pthread_join
```

That's your libnode.so, generated on your build host. See how it's importing GLIBC_2.34 symbols?

Then we check this: https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md

.NET 8 is technically supported on Ubuntu 20.04 (Focal). Which ships with GLIBC 2.31. So, those symbols are probably going to not be found there.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: Implemented GetSymbol and pointed there. This logic duplicates dotnet/runtime.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: I was wrong. .NET 8 still declares it supports Ubuntu 20.04, which ships with GLIBC 2.31. libdl symbols were moved in glibc 2.34.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: I implemented the error handling. However, I am not adding messages to the exceptions. I think I'll add those later. But they're of limited use as far as I can see: this code path is only used in the new fallback resolver for Framework, which is preempted by a File.Exists check. And we also would be shipping those libraries. So, pretty limited usage. But I'll add em anyways.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: NativeLibrary in dotnet/runtime does this through a LoadFromFilePath method that takes a 'throwOnError' argument. Which I added.

Directory.Packages.props
wasabii: > Why do we need the _._ file in the lib\netstandard2.0 folder?

nuget.org's interface determines what platforms the package is compatible with by the existence of the lib/ folders. They don't actually need items in them, but they need to be existent folders in the eventual zip file. I generally build nupkg files using MSBuild. And MSBuild doesn't allow ItemGroups with no real items. So, I just use empty files to make sure the folders exist.

Check SkiaSharp as an example: https://nuget.info/packages/SkiaSharp.NativeAssets.Win32/3.118.0-preview.2.3    They use empty folders.

> Does the buildTransitive\net461 apply to all .Net Framework version or only to 4.61?

net461 and above. This can probably be updated to net462 at this point. When I last checked net461 was the lowest supported version. Looks like 462 is now the lowest supported version.

vmoroz: > .NET 8 is technically supported on Ubuntu 20.04 (Focal). Which ships with GLIBC 2.31. So, those symbols are probably going to not be found there.

My build is currently uses Ubuntu 22.04 (it is currently the ubuntu-latest). Should I switch it to Ubuntu 20.04 to address the issue with the GLIBC 2.31 dependency?

wasabii: >Should I switch it to Ubuntu 20.04 to address the issue with the GLIBC 2.31 dependency?

I think you probably should. It seems more likely to work on all the operating systems still supported by .NET 8.

vmoroz: Could you help me understand your NuGet package contents?
- Why do we need the `_._` file in the `lib\netstandard2.0` folder?
- Does the `buildTransitive\net461` apply to all .Net Framework version or only to 4.61?

wasabii: > > Why do we need the _._ file in the lib\netstandard2.0 folder?
> 
> nuget.org's interface determines what platforms the package is compatible with by the existence of the lib/ folders. They don't actually need items in them, but they need to be existent folders in the eventual zip file. I generally build nupkg files using MSBuild. And MSBuild doesn't allow ItemGroups with no real items. So, I just use empty files to make sure the folders exist. This history is because NuGet's lib folder used to be required: because it was expected to contain an actual .NET assembly. These, being native-lib-only packages, don't.
> 
> Check SkiaSharp as an example: https://nuget.info/packages/SkiaSharp.NativeAssets.Win32/3.118.0-preview.2.3 They use empty folders.
> 
> > Does the buildTransitive\net461 apply to all .Net Framework version or only to 4.61?
> 
> net461 and above.

wasabii: Oh, another weird snippet of knowledge: there is a "bug" somewhere deep in the parsing of nuspec files that changes how the eventual zip is created based on whether or not the file does or does not have an extension. If it does not have an extension, nuget thinks it's a directory, and puts a directory in the zip file. So, if you try to put a "." file, or an "_" file in it, it actually thinks that's a directory destination, and so it ends up creating dir/_/_ or some such. Hence why I do _._.

wasabii: Oh, yeah, old common practice: https://stackoverflow.com/questions/36338052/what-do-files-mean-in-nuget-packages

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: Messages added.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -4,6 +4,7 @@
 #if !NET7_0_OR_GREATER
 
 using System;
+using System.ComponentModel;
 using System.Runtime.InteropServices;
 #if !(NETFRAMEWORK || NETSTANDARD)
 using SysNativeLibrary = System.Runtime.InteropServices.NativeLibrary;
```
wasabii, line 10: @vmoroz Under what conditions is this path active?

Directory.Packages.props
vmoroz: Thanks! I saw that without the `_._` file in the `lib\netstandard2.0` I am receiving a Nuget package warning that it had to use the .Net Framework and the package may not function correctly. I have tested the new Nuget package prototype with your PR and it all seems just fine. Hopefully we should have a new package version by the end of the day today.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -68,25 +115,77 @@ public static nint GetExport(nint handle, string name)
...
+        }
+    }
+
+    [DllImport("libdl", EntryPoint = "dlerror")]
+    private static extern nint dlerror1();
+
+    [DllImport("libdl.so.2", EntryPoint = "dlerror")]
+    private static extern nint dlerror2();
+
+    private static nint dlopen(string fileName, int flags)
```
vmoroz, line 188: I am getting a warning on the line 35 above where we pass `default` as the `fileName`. Making the first parameter nullable fixes the issue: `string? fileName`.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -101,10 +200,30 @@ private static nint dlopen(nint fileName, int flags)
...
+            return dlsym2(handle, name);
+        }
+        catch (DllNotFoundException)
+        {
+            return dlsym1(handle, name);
+        }
+    }
+
+    [DllImport("libdl", EntryPoint = "dlsym")]
+    private static extern nint dlsym1(nint fileName, string flags);
```
vmoroz, line 223: nit: the parameter names seems to be a bit incorrect: they should read `nint handle, string name`.

wasabii: Fixed.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: True. Fixed.

src/NodeApi/Runtime/NativeLibrary.cs
wasabii: So I did some changes here.

I added three versions: libc, libdl and libdl.so.2. We try them in that order.

This should align with my understanding: most modern Unix OS's have these functions in libc. That includes OS X. But also most modern Linux versions (>=GLIBC 2.34). So, this first try should get everything important.

Linux OS's <2.34 should have the functions in libdl. But those are getting old at this point. They won't be around much longer.

And then the third: libdl.so.2. My understanding is that there isn't anything that should hit this condition. But without digging into a bunch of old Linux distros I can't really tell. I think that anything that requires you to use libdl.so.2 would have been > 2.34 anyways, at which time the function moved to libc anyways. So, I think, this condition should not be hit. But I'm a bit doubtful of that.

Tests on Linux need to be run. But for whatever reason I can't run the with the testing Framework as it's set up right now. And I need some old distros.

src/NodeApi/Runtime/NativeLibrary.cs
vmoroz: No, I do not see it active under any condition. We just recently removed support for .Net 6.0 and I guess we just did not remove it here as a part of that change. Feel free to remove it.

src/NodeApi/Runtime/NativeLibrary.cs
vmoroz: BTW, I cannot compile the libnode.so on the Ubuntu 20.04. This image is absent from our blessed by secure build pipeline set of images. So, I have to continue using the Ubuntu 22.04.
Our project is not part of the official .Net runtime and I guess it is OK to say that the required minimum is the Ubuntu 22.04 since the Ubuntu 24.04 is out for some time already and the Ubuntu 22.04 is the official ubuntu-latest image on ADO and GHA.
I hope it does not cause any issues for your project.

For Mac I am going to continue to build on macos-14 since it is also the macos-latest (while .Net 8.0 supports macos-13).

wasabii: Yeah. Unfortunately, then the only remaining option would be difficult for you: building a sysroot.

vmoroz: This is a small Windows Console app to demo how to cache the function specific for the current platform.
The function records the pointer to a succeeded function and then reuses it for future calls.
This principle can be applied to your code with minimal changes:
```C#

using System.Runtime.InteropServices;

Console.WriteLine($"Hello, Module: {Handles.GetModuleHandle(null)}!");
Console.WriteLine($"Hello, Module: {Handles.GetModuleHandle(null)}!");

public static class Handles
{
    private delegate nint GetModuleHandleDelegate(string? moduleName);
    private static GetModuleHandleDelegate? s_getModuleHandle;

    public static nint GetModuleHandle(string? moduleName)
    {
        if (s_getModuleHandle is not null)
        {
            return s_getModuleHandle(moduleName);
        }

        nint result = default;
        try
        {
            result = GetModuleHandle0(moduleName);
            s_getModuleHandle = GetModuleHandle0;
        }
        catch (DllNotFoundException)
        {
            result = GetModuleHandle1(moduleName);
            s_getModuleHandle = GetModuleHandle1;
        }
        return result;
    }

    [DllImport("kernel32_bad_dll", SetLastError = true, EntryPoint = "GetModuleHandle")]
    private static extern nint GetModuleHandle0(string? moduleName);

    [DllImport("kernel32", SetLastError = true, EntryPoint = "GetModuleHandle")]
    private static extern nint GetModuleHandle1(string? moduleName);
}
```

wasabii: Fixed.

src/NodeApi/Runtime/NativeLibrary.cs
```diff
@@ -1,13 +1,12 @@
...
-#if !NET7_0_OR_GREATER
+#if !NETCOREAPP3_0_OR_GREATER
 
 using System;
+using System.ComponentModel;
 using System.Runtime.InteropServices;
-#if !(NETFRAMEWORK || NETSTANDARD)
-using SysNativeLibrary = System.Runtime.InteropServices.NativeLibrary;
-#endif
+using System.Runtime.InteropServices.ComTypes;
```
vmoroz, line 9: nit: is it being used? ` System.Runtime.InteropServices.ComTypes`

wasabii: No. VS keeps adding it when I copy and paste... something. And I forgot this one time to remove it. =/

vmoroz: Ouch... :)

Directory.Packages.props
vmoroz: It takes a bit longer to produce the package - had to restart the PR build a few times. Current ETA is ~7 hours from now.

wasabii: I see. Trying it out.

One thing I notice is that the ItemGroup is being added in the build targets file outside of a Target itself. The reason I used the `Target` was so that the item wasn't part of the `None` item group during the MSBuild evaluation pass, and thus would not appear in Visual Studio, etc.

For instance, now, when I have your package added.... including if your package is a deep transitive depedency, my Visual Studio shows three native items in the Solution Explorer: libnode.dll libnode.dylib and libnode.so.

wasabii: Other than that, update to .203 pushed. So this sub conversation should be resolved.

vmoroz: OK, I did not pay attention to it. I will do the change next week.

vmoroz: BTW, the new Nuget RID-specific packages can be also consumed on their own.
E.g. if the app is targeting only Linux platforms, then it can just set the dependency on the "linux-x64" package for x64 CPUs.

vmoroz:
The PR validation fails the formatter. You can run the following script to fix some of the issues automatically.
```
dotnet format --no-restore --severity info --verbosity detailed 
```

Directory.Packages.props
wasabii: For CoreCLR that's handled automatically based on how they publish the app they are building. It only selects the versions of the native libraries that are available that are required for their target RID.

`dotnet publish myproject.csproj -r linux-x64`, for instance, will only included the libraries in linux-x64.

That's sort of the expected flow of all this. You don't want to force your users to really make that decision.

Except for Framework where it's pretty much busted and we have to do it ourselves.

wasabii:
> @wasabii , the code formatter fails the PR validation. You can run the following script to fix some of the issues automatically.
> 
> ```
> dotnet format --no-restore --severity info --verbosity detailed 
> ```

That command just runs a bit then throws an exception. Unhandled exception: System.Exception: source text did not have an identifiable encoding

Anyway to identify the failures from the verification stuff above? Don't really see any....

vmoroz:
> > @wasabii , the code formatter fails the PR validation. You can run the following script to fix some of the issues automatically.
> > ```
> > dotnet format --no-restore --severity info --verbosity detailed 
> > ```
> 
> That command just runs a bit then throws an exception. Unhandled exception: System.Exception: source text did not have an identifiable encoding
> 
> Anyway to identify the failures from the verification stuff above? Don't really see any....

I wonder if it is related to Windows platform somehow. Last time I did not see any issues.
If you do not mind, then I can fix it and update your PR.

wasabii:
It runs fine on Linux. But modifiers no files. Feel free to apply what you need to.

vmoroz:
> I just read your edit. Hehe. Gotcha.

I see the same behavior on Windows. Anyway, the CodeQL is passing. I will play with other issues.
The CodeQL issue was related to the appearance of the new version of .Net 9.0.2 and .Net 8.0.13.
My next step is to install them locally and see if the formatter will make any difference.

vmoroz:
The 9.0.2 throws the same exception as you reported above. 
This issue is reported here https://github.com/dotnet/sdk/issues/46780 against dotnet 9.0.2.
No solution yet.

vmoroz:
This commit has all the required formatting issues: https://github.com/vmoroz/node-api-dotnet/commit/8587002a01e8b46de0c824f50b063135bf78bf7f
I cannot update this PR because you disabled the contributions from other developers.

To fix the issue I had to switch the encoding of the files to CRLF for the formatter to run.
It seems that 9.0.2 developers never tested it for LF line ending files.

wasabii:
I pulled that commit in.

vmoroz (approved):

vmoroz:
@wasabii , @jasongin , it seems I screwed up the NuGet release.
The right version that contains this PR is 0.9.5.
Though previously the build pipeline published version 0.9.6 which pretty much the same as the 0.9.4 due to my mistake.
I have unlisted the 0.9.6 from Nuget.org, but it may still cause issues in the short run.

Issue #420, opened 2025-01-16
Error related to global.node_api_dotnet.require when trying to use streams


fstanis:
I tried a simple example to test out using streams:

```
[JSExport]
public class Class1
{
    public static Stream HelloStream(string greeter)
    {
        string message = "hello " + greeter;
        byte[] messageBytes = Encoding.UTF8.GetBytes(message);
        MemoryStream stream = new MemoryStream(messageBytes);
        stream.Position = 0;
        return stream;
    }
}
```

Oddly, this throws an error:

```
const MyModule = require('./MyModule/bin/MyModule');
const stream = MyModule.Class1.helloStream('world'); // throws
```

This is the error:

```
Error: The require function was not found on the global node_api_dotnet object. Set `global.node_api_dotnet.require` before loading the module.
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.get_RequireFunction() + 0x35f
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.<>c__DisplayClass51_0.<Import>b__0(ValueTuple`2 _) + 0x1c4
    at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey, Func`2) + 0xdf
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.Import(String, String, Boolean) + 0xd7
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.<>c__DisplayClass51_0.<Import>b__0(ValueTuple`2 _) + 0x33e
    at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey, Func`2) + 0xdf
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.Import(String, String, Boolean) + 0xd7
    at Microsoft.JavaScript.NodeApi.Interop.NodeStream.CreateProxy(Stream) + 0x5a
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.<>c__DisplayClass37_0`1.<GetOrCreateObjectWrapper>b__0() + 0x43
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.GetOrCreateObjectWrapper[T](T, Func`1) + 0xd6
    at Microsoft.JavaScript.NodeApi.Interop.JSRuntimeContext.GetOrCreateObjectWrapper[T](T) + 0x8d
[...]

PR #431, opened 2025-03-04
Update streams.md to explain AOT modules

fstanis:
Explains https://github.com/microsoft/node-api-dotnet/issues/420


jasongin:
This shouldn't be required, though I understand it is currently; I want to investigate more before we accept this doc PR.

Issue #456, opened 2025-09-13
The double[][] in .NET is spitting out bad conversion code.


hetaoos:
The function name incorrectly includes []:

```cs
	private static Microsoft.JavaScript.NodeApi.JSValue from_System_Double[]_Array(double[][] value)
	{
		Microsoft.JavaScript.NodeApi.JSArray jsArray = new Microsoft.JavaScript.NodeApi.JSArray(value.Length);
		jsArray.CopyFrom<double[]>(value, 0, (double[] value) =>
		{
			jsArray = new Microsoft.JavaScript.NodeApi.JSArray(value.Length);
			jsArray.CopyFrom<double>(value, 0, (double value) =>
			{
				return (Microsoft.JavaScript.NodeApi.JSValue)value;
			});
			return (Microsoft.JavaScript.NodeApi.JSValue)jsArray;
		});
		return (Microsoft.JavaScript.NodeApi.JSValue)jsArray;
	}

	private static double[][] to_System_Double[]_Array(Microsoft.JavaScript.NodeApi.JSValue value)
	{
		Microsoft.JavaScript.NodeApi.JSArray jsArray = (Microsoft.JavaScript.NodeApi.JSArray)value;
		double[][] array = new double[][jsArray.Length];
		jsArray.CopyTo<double[]>(array, 0, (Microsoft.JavaScript.NodeApi.JSValue value) =>
		{
			jsArray = (Microsoft.JavaScript.NodeApi.JSArray)value;
			array = new double[jsArray.Length];
			jsArray.CopyTo<double>(array, 0, (Microsoft.JavaScript.NodeApi.JSValue value) =>
			{
				return (double)value;
			});
			return array;
		});
		return array;
	}
```

```xml
	<ItemGroup>
		<PackageReference Include="Microsoft.JavaScript.NodeApi" Version="0.9.15-*" />
		<PackageReference Include="Microsoft.JavaScript.NodeApi.Generator" Version="0.9.15-*" />
	</ItemGroup>
```

bhalderman-bv:
Is this issue just for the name of the function, or for the body as well? Since I was getting build errors trying to use `double[][]`, I switched to `List<List<double>>`, which fixes the method name but returns an infinitely recursive array. I was expecting output something like `[[0,1]]` and instead got `[0,1,[...]]` where the item at index 2 is a reference to the outer array. Looking at the generated code, it's nearly identical to what you have above, so I'm assuming the conversion to `double[][]` will have the same problem.

PR #457, opened 2025-09-17
Support for two-dimensional arrays such as double[][]

hetaoos:
#456


jasongin:
@hetaoos Can you add some test cases for this?

PR #476, opened 2026-05-19
Add support for .NET 10

dotnetjunky:
Attempt to add support for .NET. I used Github Copilot to ask it to generate the changes based on the previous commit to add support for .NET 9.


dotnetjunky:
The build is failing because the "dotnet build" step is trying to download the nuget package "Microsoft.NET.ILLink.Tasks" from the react-native-public feed, but the .nupkg file requires authentication.

dotnetjunky:
@jasongin @vmoroz  Could you please take a look?

.npmrc
```diff
@@ -4,5 +4,4 @@
 # Use ADO public NPM registry to be complient
 # Additional registries should be added as upstreams instead of being added here.
 
-registry=https://pkgs.dev.azure.com/ms/react-native/_packaging/react-native-public/npm/registry/
-always-auth=true
```
vmoroz, line 8: why is it changed?

src/NodeApi/Runtime/PooledBuffer.cs
```diff
@@ -52,11 +52,8 @@ private PooledBuffer(ArrayPool<byte> pool, int length, int bufferMinimumLength)
 
     public void Dispose()
```
vmoroz, line 53: why this change is needed?

docs/scenarios/js-dotnet-module.md
```diff
@@ -86,7 +86,7 @@ For a minimal example of this scenario, see
     const dotnet = require('node-api-dotnet/net8.0');
     ```
     :::
-   Currently the supported target frameworks are `net472`, `net8.0`, and `net9.0`.
+   Currently the supported target frameworks are `net472`, `net8.0`, `net9.0`, and `net10.0`.
```
vmoroz, line 89: The comment says that we support net9.0, but I see it being removed from the build matrix in the pipelines.

src/NodeApi/Runtime/PooledBuffer.cs
jasongin: Looks like it's using the new C# syntax for null-conditional method calls. Maybe there's a new analyzer that suggests it?

It's fine with me though I wouldn't have thought it was really necessary.

vmoroz:
@dotnetjunky , it is not clear from the change the main purpose of the PR.
Are you adding support for .Net 10 or replacing .Net 9 with .Net 10?
Please carefully review the changes to the pipeline files.

vmoroz:
~~I will update the NuGet feed to supply missing packages.~~
Never mind - it seems that you already updated them.

src/NodeApi/Runtime/PooledBuffer.cs
dotnetjunky: Yeah, I got a build error without that change.

.npmrc
dotnetjunky: I got a warning saying that `always-auth` has been deprecated. It's basically a no-op. So I just deleted it.

docs/scenarios/js-dotnet-module.md
dotnetjunky: Where in the build matrix that net9.0 was removed? Could you point out the file. There're a few places where we build the project with the latest framework, so I replaced .net 9 with .net 10. But other than those, I just added the new net10 to the build matrix.

dotnetjunky:
> @dotnetjunky , it is not clear from the change the main purpose of the PR. Are you adding support for .Net 10 or replacing .Net 9 with .Net 10? Please carefully review the changes to the pipeline files.

@vmoroz  The goal is to add support for .NET 10, not replacing .NET 9. There's a few places where we only build the project with the latest framework, so I replaced v9 with v10. Please point out which file I did it wrong.

docs/scenarios/js-dotnet-module.md
vmoroz: I am referring to the `publish.yml` file. This ADO pipeline is responsible for creating the NuGet files.

jasongin: I am OK with that. There are probably few if any teams on .NET 9 who haven't already migrated to .NET 10. So it's OK to exclude 9 from the build matrix until support is dropped. (Official support for .NET 9 ends Nov 2026.) Anyway it is very unlikely that there would be any bug on .NET 9 that doesn't also repro on either 8 or 10.

jasongin: Oh, it looks like in publish.yml both .NET 9 and 10 are published for Windows `win_x64` platform, but only .NET 10 for linux & mac platforms. That's odd.

vmoroz: Yep, I am for consistency. If we want to drop 9.0 then it is fine. Let's just make it consistent through the code. Though, I would rather keep 9.0 at least until the end of the official .Net support.

dotnetjunky: @vmoroz @jasongin With the current code, I saw that on windows we built net8, net9 and net47, but only built the latest net9 on macos and linux. So I thought for those two platforms, I just needed to replace net9 and net10. 

I'm happy to add net9 back for macos and linux. Let me know what you think? What's the reason we don't build net8 on those platforms?

dotnetjunky: I added .net9 back for macos and linux

vmoroz: Sounds good! .Net started to support AOT for MacOS only from version 9.0.
Thus, the full Node-API for .Net testing suite can run there starting with version 9.0.
I am not sure about Linux, I guess the reason was that we added support for that platform only after MacOS in our project. Or maybe because 8.0 must run on earlier versions of Linux and I did not know how to setup Docker at that time.

vmoroz (approved):
LGTM

docs/scenarios/js-dotnet-module.md
dotnetjunky: Ah I see. That makes sense. Thanks for the explanation.

jasongin (approved):

PR #478, opened 2026-05-27
Fix publish and release pipelines

vmoroz:
# Fix publish and release pipelines

Two unrelated breakages were blocking releases.

## publish.yml — ESRP rejects `.node` files

`Microsoft.JavaScript.NodeApi.node` is a native module — a PE/COFF DLL
with a non-`.dll` extension. ESRP's PendingAnalysis stage cannot handle
that extension and fails after a 30-minute hang.

**Fix:** rename `.node` → `.dll` before ESRP, sign it, and rename back after.

Also bumped `EsrpCodeSigning@5` → `@6` (two call sites).

## release.yml — PAT-based service connections no longer permitted

The `Nuget - ms/react-native-public` and `Npm - ms/react-native` service
connections authenticate via PAT, which policy now forbids. Managed
Identity is mandatory.

**Fix:** before each publish, an `AzureCLI@2` step fetches an AAD token
for the AzDO resource (`499b84ac-1321-427f-aa17-267ca6975798`) under MI
and overrides the service connection's auth at runtime via
`##vso[task.setendpoint id=<guid>;field=authParameter;key=apitoken]`.
`1ES.PublishNuGet@1` and `npmAuthenticate@0` then run unchanged and
transparently use the MI-derived token.

Notes:

- **Why the inline override?** ADO does not allow MI-based auth on
  service connections, so the connection itself still has to be
  configured with a PAT. The override-at-runtime pattern is the only
  way to actually publish under MI today; the stored PAT is bypassed
  on every build. This is the long-term shape unless ADO changes.
- **Temporary MI subscription.** `Office-Hermes-Windows-Bot` is used
  because the official `Office-Node-Api-DotNet-Bot` MI does not yet
  have the right permissions on these feeds. Switch back once it does.

## Drive-by

- `release.yml`: renamed `CustomPipelineTemplates` →
  `OfficePipelineTemplates` and dropped the
  `ES365AIMigrationTooling-BulkMigrated-Release` legacy build tag.

## Test runs

- Publish: <https://dev.azure.com/office/ISS/_build/results?buildId=49400440&view=results>
- Release: <https://dev.azure.com/office/ISS/_build/results?buildId=49406519&view=results>

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -386,7 +386,19 @@ extends:
                 displayName: Show RID list
 
               - ${{ if ne(variables.DisableSigning, true) }}:
-                - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@5
+                # ESRP's PendingAnalysis stage rejects the .node extension, so rename to .dll
+                # for signing and rename back after. The signing task below picks it up via the
+                # **/Microsoft.JavaScript.NodeApi.dll glob.
+                - task: PowerShell@2
+                  displayName: Rename .node to .dll for ESRP signing
+                  inputs:
+                    targetType: inline
+                    script: |
+                      Rename-Item `
+                        -Path "$(Build.SourcesDirectory)/out/bin/Release/NodeApi/aot/win-x64/publish/Microsoft.JavaScript.NodeApi.node" `
+                        -NewName "Microsoft.JavaScript.NodeApi.dll"
+
+                - task: EsrpCodeSigning@6
                   displayName: CodeSign Binaries
                   inputs:
                     ConnectedServiceName: 'ESRP-JSHost3'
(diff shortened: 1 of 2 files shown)


.ado/publish.yml
```diff
@@ -386,7 +386,19 @@ extends:
...
+                # for signing and rename back after. The signing task below picks it up via the
+                # **/Microsoft.JavaScript.NodeApi.dll glob.
+                - task: PowerShell@2
+                  displayName: Rename .node to .dll for ESRP signing
+                  inputs:
+                    targetType: inline
+                    script: |
+                      Rename-Item `
+                        -Path "$(Build.SourcesDirectory)/out/bin/Release/NodeApi/aot/win-x64/publish/Microsoft.JavaScript.NodeApi.node" `
+                        -NewName "Microsoft.JavaScript.NodeApi.dll"
```
jasongin: There is already another file named `Microsoft.JavaScript.NodeApi.dll` in this directory - it is the managed assembly DLL that the unmanaged `.node` file loads after boostrapping .NET. So it looks like this rename will overwrite that file and the final build output will not be correct.

jasongin (changes requested):

.ado/publish.yml
vmoroz: Good catch! I will fix it.

vmoroz: @jasongin , it is fixed. The test run was successful. I examined the created artifacts. It looks correct: https://dev.azure.com/office/ISS/_build/results?buildId=49440868&view=artifacts&pathAsName=false&type=publishedArtifacts

jasongin (approved):

PR #482, opened 2026-07-14
Write explicit null terminator in Utf8StringArray

### LINKED ISSUE #481: Embedding args occasionally get trailing garbage on .NET (missing null terminator on a pooled buffer)
Passing `Args` to `NodeEmbeddingPlatformSettings` intermittently fails with

```
bad option: --disable-wasm-trap-handler<random leftover bytes>
```

The tail is different every time and looks like fragments of unrelated strings from elsewhere in the process (table names, event names, and so on). It works if you create the platform right at startup, but once the app has been running for a while it starts failing, which pointed at the option value picking up stale memory.

Versions: `Microsoft.JavaScript.NodeApi` 0.9.11 with `Microsoft.JavaScript.LibNode` 20.1800.215 (Node 20.18.0), on net8.0. Reproduced on both Windows x64 and Linux x64 — it is tied to the TFM, not the OS.

## Cause

In `src/NodeApi/Runtime/Utf8StringArray.cs`:

```csharp
#if NETFRAMEWORK || NETSTANDARD
    _stringBuffer = new byte[byteLength];                       // zeroed
#else
    _stringBuffer = ArrayPool<byte>.Shared.Rent(byteLength);    // NOT zeroed
#endif

...

offset += Encoding.UTF8.GetBytes(
              src, strings[i].Length,
              (byte*)(stringBufferPtr + offset), byteLength - offset)
          + 1; // +1 for the string Null-terminator.
```

The `+ 1` reserves a byte for the terminator, but nothing ever writes the `\0`. On the netfx/netstandard path this is harmless because `new byte[]` is zero-initialized, so the reserved byte is already 0. On the .NET path the buffer comes from `ArrayPool<byte>.Shared`, which returns it with its previous contents intact, so the terminator byte can hold any leftover value. Native node then reads each argv entry as a C string and reads past the end of the intended option into that stale data.

This also explains the timing. On a fresh process the pool is empty, so the first rent returns zeroed memory and everything works. After the app has used `ArrayPool<byte>.Shared` for anything else (JSON, IO, DB drivers, and so on), the buffer comes back dirty and the option gets a trailing tail.

## Repro

The failure can be made deterministic by dirtying the relevant pool bucket first:

```csharp
using System.Buffers;
using Microsoft.JavaScript.NodeApi.Runtime;

// byteLength for ["node","--disable-wasm-trap-handler"] is (4+1)+(27+1)=33 -> 64-byte bucket
var marker = System.Text.Encoding.ASCII.GetBytes("_LEAKED_MARKER_1234567890");
for (int i = 0; i < 64; i++)
{
    var buf = ArrayPool<byte>.Shared.Rent(33);
    for (int k = 0; k < buf.Length; k++) buf[k] = marker[k % marker.Length];
    ArrayPool<byte>.Shared.Return(buf); // returned without clearing
}

var platform = new NodeEmbeddingPlatform(new NodeEmbeddingPlatformSettings
{
    Args = new[] { "node", "--disable-wasm-trap-handler" }
});
```

produces:

```
bad option: --disable-wasm-trap-handler_MARKER_1234567890_LEAKED_MARKER
```

## Fix

Write the terminator explicitly instead of relying on the buffer being zeroed:

```csharp
int written = Encoding.UTF8.GetBytes(
    src, strings[i].Length,
    (byte*)(stringBufferPtr + offset), byteLength - offset);
((byte*)stringBufferPtr)[offset + written] = 0;
offset += written + 1;
```

Clearing the rented buffer would also work, but writing the terminator is cheaper. The pointer array (`ArrayPool<nint>.Shared.Rent`) is not affected, since it is passed with an explicit count.

This is hard to diagnose because it passes in isolation and only surfaces under load or after warm-up. I can open a PR with the terminator fix and a test that poisons the pool if that would help.

tolmachev-pravo:
Fixes #481.

`Utf8StringArray` reserved a byte for each string's null terminator (the `+ 1` in the encoding loop) but never wrote the `\0` — it relied on the reserved byte already being zero. That holds on the `NETFRAMEWORK`/`NETSTANDARD` path (`new byte[]` is zeroed), but on .NET the buffer is rented from `ArrayPool<byte>.Shared`, which is not zero-initialized. Once the pool has been used, the terminator byte can contain leftover data, and native node reads each argv entry past its end — e.g. `bad option: --disable-wasm-trap-handler<garbage>`. It only reproduces after warm-up (dirty pool), which made it hard to track down.

Fix: write the terminator explicitly.

Adds a regression test that dirties `ArrayPool<byte>.Shared` and verifies the marshaled strings round-trip exactly (fails on `main`, passes with this change). Testing the internal struct needs `InternalsVisibleTo` on the test assembly.

Verified: `dotnet test --framework net8.0 --filter Utf8StringArrayTests` passes with the fix and fails without it.

### DIFF
--- a/src/NodeApi/NodeApi.csproj
+++ b/src/NodeApi/NodeApi.csproj
@@ -28,6 +28,10 @@
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.JavaScript.NodeApi.Test" />
+  </ItemGroup>
+
   <Target Name="SetRuntimeConfigValues" BeforeTargets="GenerateBuildRuntimeConfigurationFiles">
     <ItemGroup>
       <RuntimeHostConfigurationOption Include="System.Runtime.InteropServices.EnableConsumingManagedCodeFromNativeHosting" Value="true" />
--- a/src/NodeApi/Runtime/Utf8StringArray.cs
+++ b/src/NodeApi/Runtime/Utf8StringArray.cs
@@ -46,9 +46,13 @@ internal struct Utf8StringArray : IDisposable
             fixed (char* src = strings[i])
             {
                 Utf8Strings[i] = stringBufferPtr + offset;
-                offset += Encoding.UTF8.GetBytes(
-                    src, strings[i].Length, (byte*)(stringBufferPtr + offset), byteLength - offset)
-                    + 1; // +1 for the string Null-terminator.
+                int encodedLength = Encoding.UTF8.GetBytes(
+                    src, strings[i].Length, (byte*)(stringBufferPtr + offset), byteLength - offset);
+
+                // Rented pool buffers are not zero-initialized, so write the terminator explicitly
+                // rather than relying on the reserved byte already being zero (#481).
+                *(byte*)(stringBufferPtr + offset + encodedLength) = 0;
+                offset += encodedLength + 1;
             }
         }
     }
--- a/test/Utf8StringArrayTests.cs
+++ b/test/Utf8StringArrayTests.cs
@@ -1,0 +1,71 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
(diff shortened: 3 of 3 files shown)


tolmachev-pravo:

Issue #486, opened 2026-08-03
SIGSEGV (exit 139) on Node worker_threads teardown after require('node-api-dotnet/net10.0') - pthread TLS destructor use-after-free


GalaxiasKyklos:
[gdb-repro-run.sh](https://github.com/user-attachments/files/30668296/gdb-repro-run.sh)
[gdb-worker-teardown-bt.txt](https://github.com/user-attachments/files/30668298/gdb-worker-teardown-bt.txt)
[gdb-worker-teardown-fault-mapping.txt](https://github.com/user-attachments/files/30668297/gdb-worker-teardown-fault-mapping.txt)
[package.json](https://github.com/user-attachments/files/30668295/package.json)

## Summary

Loading `node-api-dotnet` inside a Node.js **`worker_threads` Worker** and then tearing that worker down - either via `worker.terminate()` from the parent **or** a graceful `process.exit(0)` inside the worker - reliably crashes the **entire process** with a native **SIGSEGV (exit code 139)** on Linux.

The crash requires **zero interop calls**: merely `require('node-api-dotnet/net10.0')` on the worker thread is enough. The same teardown on the **main** thread is clean, and a worker that does **not** load node-api-dotnet is clean.

A native backtrace shows the fault is in **glibc's per-thread TLS-destructor cleanup** (`__nptl_deallocate_tsd`) as the worker thread exits: a `pthread_key` **destructor function pointer is dangling** (points to unmapped memory), so calling it segfaults.

This is **distinct from** the thread-safe-function teardown issue in [PR #480](https://github.com/microsoft/node-api-dotnet/pull/480): applying that patch does **not** fix this crash, and neither does removing the `JSThreadSafeFunction.Release()` call entirely.

## Environment

| | |
|---|---|
| node-api-dotnet | **0.9.21** |
| .NET runtime | **10.0.10** (Microsoft.NETCore.App), framework-dependent |
| Target framework moniker | `net10.0` |
| OS / libc | Linux x64, glibc (Debian 12 and Azure Linux 3.0 both affected) |

### Node.js version dependence (matches the PR #480 threshold)

| Node.js | Result on worker teardown |
|---|---|
| 22.17.0 | **SIGSEGV / exit 139** |
| 24.13.0 | **SIGSEGV / exit 139** |
| 24.18.1 (>= 24.14) | **Deadlock / hang** (process never exits) |

## Minimal reproduction

[...]

PR #487, opened 2026-08-03
Fix SIGSEGV when a worker_threads Worker that loaded node-api-dotnet …

GalaxiasKyklos:
## Summary

Fixes a **SIGSEGV** (process crash, exit 139) that occurs on Linux when a `worker_threads` Worker loads `node-api-dotnet` and the worker is then terminated.

Issue: #486 

## Repro

Load the module inside a Worker, then terminate the worker:

```js
// worker.cjs
const { parentPort } = require('node:worker_threads');
require('node-api-dotnet/net10.0');
parentPort.postMessage('ready');
```

```js
// main.cjs
const { Worker } = require('node:worker_threads');
const w = new Worker('./worker.cjs');
w.once('message', async () => {
  await w.terminate();
  console.log('terminated ok');
});
```

On Linux the process crashes with SIGSEGV instead of printing `terminated ok`.

## Root cause

The native host (`Microsoft.JavaScript.NodeApi.node`) is compiled with **NativeAOT**, so the `.node` embeds its own .NET runtime. That runtime registers per-thread cleanup with the OS via a `pthread_key` destructor whose function pointer points **into this module's own code**.

When a Worker environment is torn down, Node.js unloads the addon (`dlclose`) while the worker's OS thread is **still alive**. The `pthread_key` destructor is still registered, but now points at unmapped memory. When the worker thread subsequently exits, glibc's `__nptl_deallocate_tsd` invokes that dangling destructor → **SIGSEGV**.

This was confirmed with gdb: at crash time the faulting destructor belonged to the `.node`, which had **0 memory mappings** (already `dlclose`d). No-op'ing `dlclose` via `LD_PRELOAD` produced a clean exit, confirming the unmap-while-thread-alive diagnosis.

## Fix

Pin the native host module for the lifetime of the process on Linux. Resolve the module's own path via `dladdr` on one of its functions, then re-open it with `dlopen(RTLD_NOLOAD | RTLD_NODELETE)`:

- `RTLD_NOLOAD` resolves the already-loaded module without loading a second copy.
- `RTLD_NODELETE` keeps it mapped for the process lifetime, and the extra (never-released) reference prevents Node's `dlclose` from unmapping it.

This keeps the `pthread_key` destructor address valid for the life of the process. The change is:

- **Scoped to Linux/glibc** — Windows module/thread teardown does not hit this path.
- **Best-effort** — any failure is traced and non-fatal, so it never blocks module init.

## Testing

Reproduced and validated in Docker (`mcr.microsoft.com/dotnet/sdk:10.0` + Node 24.13.0, which crashes deterministically):

| | Before | After |
|---|--------|-------|
| Node 24.13.0 | exit 139 (SIGSEGV) | **exit 0** (`terminated ok`), 5/5 runs |

`NodeApi` builds clean on `net8.0`, `net9.0`, `net10.0`, and `netstandard2.0`.

> Note: This addresses only the crash. A separate hang on Node.js >= 24.14 (TSFN release deadlock during env teardown) is fixed independently.

### DIFF
--- a/src/NodeApi/DotNetHost/NativeHost.cs
+++ b/src/NodeApi/DotNetHost/NativeHost.cs
@@ -43,12 +43,93 @@ internal unsafe partial class NativeHost : IDisposable
         }
     }
 
+    private static bool s_moduleUnloadPrevented;
+
+    /// <summary>
+    /// Pins this native host module in memory so the OS never unloads it.
+    /// </summary>
+    /// <remarks>
+    /// This native host is compiled with NativeAOT, so it embeds a .NET runtime whose
+    /// per-thread cleanup is registered with the OS via a <c>pthread_key</c> destructor that
+    /// points into this module's own code. Node.js unloads (<c>dlclose</c>) an addon when the
+    /// environment that loaded it is torn down. When a <c>worker_threads</c> Worker loads this
+    /// module and is then terminated, Node unloads the module while the worker's OS thread is
+    /// still alive; the still-registered destructor then points at unmapped memory and the
+    /// process crashes with SIGSEGV as the thread exits (glibc <c>__nptl_deallocate_tsd</c>).
+    /// Keeping the module mapped for the lifetime of the process keeps that destructor valid.
+    /// <para/>
+    /// This only affects Unix (glibc) hosting; on Windows module/thread teardown does not hit
+    /// this issue. The pin is best-effort: any failure is traced but does not block init.
+    /// </remarks>
+    private static unsafe void PreventModuleUnload()
+    {
+        if (s_moduleUnloadPrevented)
+        {
+            return;
+        }
+
+        s_moduleUnloadPrevented = true;
+
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            return;
+        }
+
+        try
+        {
+            // Resolve the file path of this shared library from the address of one of its
+            // own functions, then re-open it with RTLD_NODELETE so it is never unmapped.
+            nint moduleFunction =
+                (nint)(delegate* unmanaged[Cdecl]<napi_env, napi_value, napi_value>)
+                &InitializeModule;
+
+            if (dladdr(moduleFunction, out Dl_info info) != 0 && info.dli_fname != default)
+            {
+                // RTLD_NOLOAD resolves the already-loaded module without loading a new copy;
+                // RTLD_NODELETE keeps it mapped for the process lifetime. The extra (never
+                // released) reference also prevents Node's dlclose from unmapping it.
+                const int RTLD_LAZY = 0x0001;
+                const int RTLD_NOLOAD = 0x0004;
+                const int RTLD_NODELETE = 0x1000;
+                nint handle = dlopen(info.dli_fname, RTLD_LAZY | RTLD_NOLOAD | RTLD_NODELETE);
+                Trace($"    Pinned native host module ({(handle != default ? "ok" : "no-op")}).");
+            }
+            else
+            {
+                Trace("    Could not resolve native host module path to pin it.");
+            }
+        }
+        catch (Exception ex)
+        {
+            Trace("    Failed to pin native host module: " + ex);
+        }
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    private struct Dl_info
+    {
+        public nint dli_fname;
+        public nint dli_fbase;
+        public nint dli_sname;
+        public nint dli_saddr;
+    }
+
+    [DllImport("libc.so.6")]
+    private static extern int dladdr(nint addr, out Dl_info info);
+
+    [DllImport("libc.so.6")]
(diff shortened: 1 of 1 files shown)


src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,13 +43,94 @@ public static void Trace(string msg)
...
+    private static unsafe void PreventModuleUnload()
+    {
+        if (s_moduleUnloadPrevented)
+        {
+            return;
+        }
+
+        s_moduleUnloadPrevented = true;
+
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
jasongin: You mentioned above that Windows does not hit this problem; do you know whether this issue affects Mac OS?

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,13 +43,94 @@ public static void Trace(string msg)
...
+    [StructLayout(LayoutKind.Sequential)]
+    private struct Dl_info
+    {
+        public nint dli_fname;
+        public nint dli_fbase;
+        public nint dli_sname;
+        public nint dli_saddr;
+    }
+
+    [DllImport("libc.so.6")]
```
vmoroz: Please fix the formatter issues - it says to use the `LibraryImportAttribute` instead of `DllImportAttribute`.

src/NodeApi/DotNetHost/NativeHost.cs
GalaxiasKyklos: I've extended the pin to macOS in the latest commit, I could only validate on Linux (Docker repro); the macOS path is best-effort and unvalidated on real hardware. Happy to gate it Linux-only if you'd prefer until it's validated on a Mac.

src/NodeApi/DotNetHost/NativeHost.cs
GalaxiasKyklos: Done, converted the  dladdr / dlopen  imports to source-generated [LibraryImport]  in the latest commit

src/NodeApi/DotNetHost/NativeHost.cs
jasongin: PR builds run on macOS, so at least they will validate it's not broken (once the package issues are resolved). But I'd prefer to confirm whether or not this fix is actually needed on that platform. Is there a simple repro? I use macOS as my main dev environment now, so I could easily try it out.

GalaxiasKyklos: Thank you! I put together this gist with instructions on how to repro it on macOS

https://gist.github.com/GalaxiasKyklos/8bb730adcd486af5e0067eb2f1c69055

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,13 +43,125 @@ public static void Trace(string msg)
...
+        {
+            return;
+        }
+
+        s_moduleUnloadPrevented = true;
+
+        bool isMacOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && !isMacOS)
+        {
+            return;
```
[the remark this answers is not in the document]
GalaxiasKyklos: Updated

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,13 +43,125 @@ public static void Trace(string msg)
...
+
+    // dladdr / dlopen live in libc.so.6 on Linux (glibc) and libSystem on macOS.
+    [LibraryImport("libc.so.6", EntryPoint = "dladdr")]
+    private static partial int DlAddrLinux(nint addr, out Dl_info info);
+
+    [LibraryImport("libSystem", EntryPoint = "dladdr")]
+    private static partial int DlAddrMacOS(nint addr, out Dl_info info);
+
+    [LibraryImport("libc.so.6", EntryPoint = "dlopen")]
+    private static partial nint DlOpenLinux(nint filename, int flags);
```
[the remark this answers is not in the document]
GalaxiasKyklos: Done

src/NodeApi/DotNetHost/NativeHost.cs
jasongin: I reproduced the crash on macOS, however the fix doesn't work (even after I fixed this OS condition in the code). I'll try to debug when I have more time later.

jasongin: This set of changes makes the repro code run and exit cleanly on macOS: https://github.com/microsoft/node-api-dotnet/commit/d9f5371a299a3366941e38060cde5c0f4770f10a

In addition to fixing the module-pinning to work with macOS system libraries, it fixes a cleanup ordering issue in `JSTsfnSynchronizationContext`.

